Issue #21, opened 2013-12-13
Kiosk mode


marcusramberg:
I would like to have a restricted version of Convos, which only lets you log in and join a channel, without settings or persistence. For running a demo mode and embedding in company pages.

jhthorsen:
See also https://github.com/Nordaaker/convos/issues/89

jhthorsen:
My idea is to use JWT (JSON web tokens) with enough information for a user to register and automatically join a server/chat room.

The payload of the JWT is a JSON structure with information, such as username, server and dialog/channel name. The signature is a checksum which the server can use to validate the payload. This means that unless the token was generated from a secure party, then Convos (or any JWT party) will reject the whole token.

The last part about "kiosk mode" is that the actual user interface will be stripped down. Like not having any menus. I would do that part by another query parameter, like "?kiosk=1", or maybe even a list of elements to show, like "?kiosk=participants,heading". Using such a parameter will then make convos render nicely inside an iframe.

jhthorsen:
Closing this in favour of #346. (Because of age)

PR #200, opened 2014-09-18
Kiosk mode #21

jhthorsen:
I'm thinking the "kiosk mode" should be Convos without any goto-anything and conversation list. It could however have the nick list, actions and "Convos logo" so you can see what's going on with the server connection.

I think it should be super minimalistic, suitable for being inside an iframe, embedded on another web page.

I don't think the user need to choose nick or any other settings. This is so the user can just "jump right in and chat". (Changing the nick however is OK, of course)

idea-for-kiosk-mode

TODO:
- [x] Generate nick
- [x] Generate user account
- [x] Validate "server" in URL
- [ ] Delete idle user
- [x] Auto join channel on connect
- [x] Hide elements in gui
- [ ] Disable actions when session kiosk=1

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,11 @@
+## 0.85
+- Fix input field in iOS #167
+- Add feedback when websocket could not be established #192
+- Add /profile/delete #104
+- Change to "flat" design
+- Remove Convos::Loopback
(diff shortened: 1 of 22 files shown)


jhthorsen:
Closing this because of inactivity.

PR #250, opened 2015-11-18
cpanm --installdeps needs to be told to use the current directory

pink-mist:
Missed supplying current directory to `cpanm --installdeps`


jhthorsen:
Nice catch! :+1:

PR #254, opened 2016-02-29
add md to html support

s1037989:
This implementation has problems.  What do you think about the methodology of the implementation?

- It chops off a character after converting to html.  I could use help with this.
- It supports `*` only, and not `_`.  Markdown, IIRC, supports `*` and `_` interchangeably.
- `<strong>` didn't seem to work, so I used `<b>` instead.  IIRC, `<em>` *did* work.
- Is there any other markdown type that should be supported?
- I would have created a test, but I couldn't find an existing test that tested `autoLink()` to copy from.

If this implementation methodology is acceptable, it seems that this would also be the way to handle emoticons / emoji.  What is the opinion of support for either of these?

### DIFF
--- a/assets/assetpack.db
+++ b/assets/assetpack.db
@@ -206,10 +206,10 @@ format=js
 minified=0
 mtime=1456663191
 [riot:riot/dialog-message.tag]
-checksum=80d09974bc
+checksum=76b93ef296
 format=js
 minified=0
-mtime=1456670546
+mtime=1456787877
 [riot:riot/new-dialog.tag]
 checksum=f526f50501
 format=js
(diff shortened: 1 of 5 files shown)


jhthorsen:
Tests are completely lacking in the JS side. I would greatly appreciate if someone explained to me a good way to write those tests.

I will have a look at this PR tomorrow. It's a bit too late for me now.

jhthorsen:
Thanks for the PR, but I thought the JavaScript was a bit too in effective: It does a lot of indexOf().

d71b579 supports *, _, **, __, *** and ___. I think I bundle too few versions of the font, since the font-weight 500 does not work, so I changed it to bold instead.

Let me know what you think.

PR #257, opened 2016-04-03
Made user.js thinner

jhthorsen:
This branch cannot be merged as-is.


marcusramberg:
I see this isn't ready yet, so I'll leave it alone.

jhthorsen:
I think the interesting changes are in ebc6dfb. I wonder if @jberger agrees in the direction i went :smile: 

Btw: This is all in "one-point-oh" now.

PR #258, opened 2016-04-05
#255 Update messages() to query messages between after and before

s1037989:
DO NOT MERGE YET!

So here's a brand new crack at messages().  I implemented a cursor (word stolen from databases) that allows searching thru a series of month-hashed data log files.  It always has an upper and lower bound: after X and before X.  before defaults to now, after defaults to before - 1 month.

This new routine will retrieve up to {limit} messages within the after-before bounds.

By standardizing on the search pattern to always be after-before, I was able to simplify the retrieval from disk (iow, it doesn't matter if you specify before or after, we will always read the messages with File::ReadBackwards).

I was also able to simplify the code that ends recursion.

I need to add several more tests to ensure that this is all working as expected.  I'm at a resting point for now and want to put this out for RFC right now.  :)

Ultimately, I think this is a more simple implementation of messages() and more powerful in terms of search.  I'm not yet aware of any disadvantages.

Oh, one design question: since the query is always between after and before, what should happen if the query is such that after comes after before?

    Between:   XXXXXX[after-----before]XXXXXX
    Question:   <---before]XXXXXX[after--->

Does this make sense?  How should it be handled?  Is there any reason to accept a search query for the second case?

Please let me know any comments that you have.


lib/Convos/Core/Backend/File.pm
```diff
@@ -80,13 +80,13 @@ sub messages {
...
   $re = qr/^(\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}) (.*$re.*)$/;
 
-  $args{after}  = Time::Piece->strptime($query->{after},  '%Y-%m-%dT%H:%M:%S')if $query->{after};
-  $args{before} = Time::Piece->strptime($query->{before}, '%Y-%m-%dT%H:%M:%S') if $query->{before};
-  # If neither before nor after are defined, search everything before now
-  $args{before} = gmtime if !$args{before} && !$args{after};
   $args{limit}    = $query->{limit} || 60;
   $args{messages} = [];
   $args{re}       = $re;
+  # TODO: What to do when after is after before?
```
jhthorsen: Do you mean if "before" is after "after"..?

lib/Convos/Core/Backend/File.pm
```diff
@@ -80,13 +80,13 @@ sub messages {
...
+  $args{before}   = $query->{before} ? Time::Piece->strptime($query->{before}, '%Y-%m-%dT%H:%M:%S') : gmtime;
+  $args{after}    = $query->{after}  ? Time::Piece->strptime($query->{after},  '%Y-%m-%dT%H:%M:%S') : $args{before}->add_months(-1);
```
jhthorsen: Does this mean that the default search is to search for messages the last month, instead of beginning of time? I'm not sure if I want that. I think we should rather add those limitations in the API / JavaScript code.

lib/Convos/Core/Backend/File.pm
```diff
@@ -205,23 +205,22 @@ sub _log_file {
...
-  # this code is running in a sub process, so "guard" does not have to be localised
-  return $args->{messages} if ++$self->{guard} > ($ENV{CONVOS_MAX_SEARCH_BACK_MONTHS} || 12);
-  my $file = $self->_log_file($obj, $args->{before} || $args->{after});
-  my $pindex = $args->{before} ? 0 : -1;
-  my $FH = $pindex ? IO::File->new($file, 'r') : File::ReadBackwards->new($file);
-  my $found = 0;
+  return $args->{messages} if $args->{cursor} < $args->{after};
+  my $file = $self->_log_file($obj, $args->{cursor});
+  $args->{cursor} = $args->{cursor}->add_months(-1);
+  my $FH = File::ReadBackwards->new($file);
```
jhthorsen: I like how you just use File::ReadBackwards :+1:

lib/Convos/Core/Backend/File.pm
s1037989: "before" should be a later date, i.e. we are searching for messages "before" a given date.
"after" should be an earlier date, i.e. we are searching for messages "after" a given date.

As such, after should come before, for a normal "between".  But what happens if we are supplied a query where "after" comes after "before"?

lib/Convos/Core/Backend/File.pm
s1037989: So yeah, I'm not sure what behavior you're wanting, so please let me know that.

To explain the code (since it wasn't commented...  :(

If the query does not provide "before" default to now (that was original behavior).
If the query does not provide "after" default to one month before "before" (meaning a 1-2 month time span).

My reason for implementing this strict behavior (to enforce the presence of "before" *and* "after") was to prevent deep recursion, to go from now until the beginning of time.  Instead, this puts a hard stop-gap on the recursion, but also on length of search (just be explicit about the before / after).  The other benefit to always searching between "after" and "before" is that we can focus on just one type of file reading: File::ReadBackwards

lib/Convos/Core/Backend/File.pm
jhthorsen: Yes, of course. You're right. In that case this method should return an empty array of messages. I don't you need to implement any additional check for that.

lib/Convos/Core/Backend/File.pm
jhthorsen: Should we also add a restriction on how long timespan you can search? Like for now, we could say that before and after can max be one year apart?

lib/Convos/Core/Backend/File.pm
s1037989: Perfect!  I'll check for that case with tests and ensure it happens.

lib/Convos/Core/Backend/File.pm
s1037989: That's a great question and because of the expense of opening all these files on disk it seems like restricting to a max of 12 months would be wise.  Should we make that a user-config option so that anyone controlling their own convos could override it?  I'm think about the {limit}, and how it has the ability to be overridden.  I think we should do it in the same way: hardcode to 12 but let {max_months} (or whatever) in the query override that hard-coded 12.

Or, I'm not sure where the query can get set.  Can, theoretically, the front end code control that attribute?  We wouldn't want someone to create a DoS thru a large {limit} and vast {max_months}.

jhthorsen: For now, just hard code it. We'll make it dynamic if someone asks for it.

Convos is all about clever defaults 😄

s1037989: Solid!  I've got my game plan.  Off I go...

s1037989: Solid!  I've got my game plan.  Off I go...

s1037989:
Ok, I think this patch is ready for merge.  I have more tests added.  Let me know if there are any other tests I should be checking.

**NOTE:** I did squash all the commits into one.  This shouldn't be a problem because this PR was never merged.  I did have to force push to my github fork but, again, this shouldn't be a problem.

I just wanted to state that in case anyone is aware that this could be problematic, unbeknownst to me.

I am very happy with how _messages() turned out, but I'm less thrilled with messages().  Specifically 86-100.

Check out my changes and let me know what you think, let me know if I need to add any additional comments anywhere.

lib/Convos/Core/Backend/File.pm
```diff
@@ -80,14 +80,26 @@ sub messages {
...
+  } elsif ( !$query->{before} && $query->{after} ) {
+    $args{after} = Time::Piece->strptime($query->{after}, '%Y-%m-%dT%H:%M:%S');
+    $args{before} = $args{after}->add_months(12);
+  } elsif ( $query->{before} && $query->{after} ) {
+    $args{before} = Time::Piece->strptime($query->{before}, '%Y-%m-%dT%H:%M:%S');
+    $args{after} = Time::Piece->strptime($query->{after}, '%Y-%m-%dT%H:%M:%S');
+    # return no messages if args > before or before - args > 12
+    $args{after} = $args{before}->add_months(1) if $args{before} - $args{after} > $args{before} - $args{before}->add_months(-12);
+  }
+  $args{cursor} = $args{before};
```
jhthorsen: I think the if/else/elsif above could collapse to:

```
$args{before}
  = $query->{before} ? Time::Piece->strptime($query->{before}, '%Y-%m-%dT%H:%M:%S') : gmtime;

$args{after}
  = $query->{after}
  ? Time::Piece->strptime($query->{after}, '%Y-%m-%dT%H:%M:%S')
  : $args{before}->add_months(-12);
```

It's fine that if someone just specifies "after" that they can search from "after" until "now". If we want to limit it further, we can do so in the JavaScript.

I don't get the last "return no messages if args > before or before - args > 12" statement.

The only reason for my change is that if it's easier to understand the code, it's easier to not mess up. But I'm not going to block this because you got a lot of tests, which makes it safe either way.

jhthorsen: You decide @s1037989: If you're happy, then hit the merge button 👍

s1037989: > I think the if/else/elsif above could collapse to:

This misses an important case: if "after" is supplied but not "before" I want "before" to be "after" plus 12 months.  This series of conditionals allows for that situation which allows for a search to be limited to 12 months which is both less expensive and not prone to DoS.

Since you approve of my original code I'll leave it and comment it.  Indeed, improvements can always be made later!

I'll make my comments, and merge.  Thanks for all the feedback!

jhthorsen: Thanks! 👍

s1037989:
Done!  I like messages() much better now.  It's better organized and is well commented.  Hopefully this is to your liking because it's getting merged right now!!  :)

PR #261, opened 2016-06-10
Prevent confidential data from leaking in whois

stefanmb:
Hi folks,

I just started using Convos today and I've never used Perl, so bear :bear: with me, but I noticed a potential security issue.

I'm running Convos behind nginx as a reverse proxy and SSL endpoint with basic auth.

In the Convos code the whois `realname` field appears to be populated with the full URI where the Convos instance is running. When using basic auth this means `user:password@domain.tld` is displayed in the whois info. This seems like a very bad thing, I patched it quickly by just duplicating the user string in the real name field.

Is there a better way to fix this?

Thanks!

### DIFF
--- a/lib/Convos/Core/Connection.pm
+++ b/lib/Convos/Core/Connection.pm
@@ -486,7 +486,7 @@ sub irc_rpl_whoisuser {
   my $nick   = $params->[1];
 
   $self->{whois}{$nick}{host}     = $params->[3];
-  $self->{whois}{$nick}{realname} = $params->[5];
+  $self->{whois}{$nick}{realname} = $params->[2];
   $self->{whois}{$nick}{user}     = $params->[2];
 }
 


jhthorsen:
Hi. Why are you using basic auth? Convos has it's own login mechanism...

stefanmb:
@jhthorsen Thanks for getting back to me so quickly! Convos does have its own login mechanism, but by default it allows arbitrary users to register. I do realize there is an environment variable for a shared secret necessary to register new users, but I felt it's more prudent to rely on nginx's security to make sure no unauthorized use occurs. 

Irrespective of my use case, I'm not sure why the URI is exposed as the real name field. Even ignoring the basic auth issue it doesn't really seem to serve a purpose other than to advertise where the instance of Convos is being hosted to everyone.

jhthorsen:
Sorry, but I can't follow up a PR based on "I felt...". Please provide facts that the current login method does not work as expected.

The reason for exposing the URL is exactly that: So other convos instances can notice each other. Unfortunately this has not yet been taken to the next step, so at this point it doesn't help out much.

stefanmb:
@jhthorsen I do not know of any bugs with the current login method in Convos. Given the size of nginx's userbase versus Convos' userbase I prefer to rely on the security offered by nginx versus exposing my Convos Docker container directly to the public. In my opinion showing the full URI potentially containing passwords and usernames is a security flaw. 

However, I will close this PR now, thank you for your time and comments.

jhthorsen:
Both the session based security in Convos and basic auth both require SSL. If you don't use HTTPS (tls/ssl) then none of the security models are secure: They both suffer from the "man in the middle" attack. The server in this case does not help you with that. Other than that, Mojolicious (the web framework Convos is built on) has a very secure session handling.

One thing we might have done is setting a random invite code by default, to prevent opening Convos to the world.

stefanmb:
I understand, that makes sense. My use case is this: I don't want to deal with having to set up Redis, Mojolicious, etc. Instead I simply deployed the Docker container you provide behind nginx. In my case nginx provides SSL and reverse-proxies to the Convos container on the same host. I may revise my setup and use the shared secret instead of basic auth. Thanks again!

jberger:
https://github.com/kraih/mojo/commit/70c09a58aa22bf6670a36e35e5e2d861260a6171 now prevents the username and password from being rendered when the url is stringified. Therefore this should no longer be a problem (I think!)

stefanmb:
@jberger Sweet! I'll see about upgrading and getting rid of my floating patch. Thanks!

jberger:
Note that that version of mojolicious hasn't been released yet, it is part of the 7.0 release changeset.

PR #262, opened 2016-08-13
convert to computed properties from data initializer in info

jberger:
This is untried and untested but I think it might help.


jhthorsen:
The *issue* with computed values is that it will change every time the participant list change instead of showing a snapshot of when you asked for the information.

Not sure what is the best user experience.

jberger:
Ah interesting point. When i get back i need to figure out how to deploy
and work on a live instance

jhthorsen:
Have you tried ```./script/convos install``` or just ```cpanm --installdeps .``` ?

PR #264, opened 2016-08-17
Api over WebSockets, instead of REST

jhthorsen:
This PR is not ready for merging, but I'm opening it anyway so we can discuss https://github.com/Nordaaker/convos/blob/0b6720245faea0a47c70ef3cd21cf19aee9d20e0/lib/Convos/Plugin/StructuredWS.pm

Does anyone have another idea about how we can communicate over websockets?

So what I want to accomplish is that most (not all - more on that later) of the application will call remote methods over websockets, instead of using REST. I simply don't see why we won't utilize the connection we already have open.

Not all: So... I think there are some methods that need to run as standard HTTP requests: login, register and logout. The reason for that is that I don't know how to set the session Cookie, without using standard GET/POST requests.

I've looked into WAMP/autobahn, JSON-RPC, and @jberger's multiplex plugin, but I find them too low level without adding any value for the RPC case I'm trying to solve now.

Please correct me if I'm wrong.


jhthorsen:
So we've discussed this and come to the conclusion that this PR is not going to go forward. Instead we will use plain HTTP/REST and OpenAPI and WebSockets only for sending messages and receiving events from the backend.

PR #275, opened 2016-09-12
Added BotServ and HostServ short commands

tinaira:
Added BotServ and HostServ short commands


jhthorsen:
I'm a bit "scared" of adding too many shortcuts, since there might be a time we have too many...

Anyhow: We will deal with that later.

Thanks for the PR! 👍

PR #278, opened 2016-09-13
Adding Operator command identify

tinaira:
Now you will be able to identify you'r self as a operator now on.

### DIFF
--- a/assets/js/core/commands.js
+++ b/assets/js/core/commands.js
@@ -61,6 +61,12 @@ Convos.commands = [
     example: "/kick <nick>"
   },
   {
+    command: "oper",
+    aliases: ["oper"],
+    description: "Verify your self as operator of the network.",
+    example: "/oper operid operpassword"
+  },
+  {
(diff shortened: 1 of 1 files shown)


jhthorsen:
I don't think this will work, since the "OPER" command is not supported in the backend.

jhthorsen:
https://github.com/Nordaaker/convos/blob/master/lib/Convos/Core/Connection/Irc.pm#L164

tinaira:
@jhthorsen ou damn :a:

jhthorsen:
Unfortunately, it's not that simple. I will look at this evening.

tinaira:
Thank you

jhthorsen:
@tinaira: Could you check if the OPER command is supported now? There's no autocomplete, but it should be supported in version 0.99_24 (62ec62285927787b263396a12b99e76fb001cbdc)

jhthorsen:
I'm closing this because it has been inactive for quite some time.

PR #280, opened 2016-09-17
Convos::Plugin::ShareDialog - For sharing IRC logs

jhthorsen:
This PR feature the very first optional plugin for Convos, which allow you to share a dialog with people who does not have a Convos account.

Ideas for later:

* Add websocket support, so a visitor can see that chat in real time, as someone writes message.


lib/Convos.pm
```diff
@@ -150,6 +142,18 @@ sub _home_relative_to_lib {
...
+sub _plugins {
+  my $self    = shift;
+  my $plugins = $self->config('plugins');
+
+  unshift @{$self->plugins->namespaces}, 'Convos::Plugin';
+
+  $ENV{CONVOS_PLUGINS} //= $plugins ? '' : 'Sharelog';
+  $plugins ||= {};
+  $plugins->{$_} = {} for split /,/, $ENV{CONVOS_PLUGINS};
+  $self->plugin($_ => $plugins->{$_}) for grep { $plugins->{$_} } keys %$plugins;
```
jhthorsen: I think this is a much better idea, that plugins are attached to the $app object, instead of $core.

assets/vue/convos-dialog-settings.vue
```diff
@@ -25,6 +25,14 @@
...
         </div>
       </div>
+      <div class="row" v-if="settings.sharelog_enabled">
+        <div class="col s12 li-link">
+          <a href="#" target="_blank">
+            <i class="material-icons">share</i>
+            Share conversation
+          </a>
+        </div>
+      </div>
```
jhthorsen: I wonder how this can be solved smoothly with plugins. Can they modify the page, without actually having to modify the vue templates?

lib/Convos.pm
```diff
@@ -129,9 +113,17 @@ sub _config {
...
+  $config->{secure_cookies}    ||= $ENV{CONVOS_SECURE_COOKIES}    || 0;
+
+  # public settings
+  $config->{settings} = {
+    contact           => $config->{contact},
+    default_server    => $config->{default_server},
+    forced_irc_server => $config->{forced_irc_server} ? true : false,
+    invite_code       => $config->{invite_code} ? true : false,
+    organization_name => $config->{organization_name},
+  };
```
jhthorsen: It's much easier for plugins to change the public settings, when it's a variable instead of a helper.

lib/Convos/Plugin.pm
```diff
@@ -0,0 +1,39 @@
+package Convos::Plugin;
```
jhthorsen: Is there really any reason for this..? I think it's cleaner than inheriting directly from Mojolicious::Plugin.

lib/Convos.pm
jberger: 👍

lib/Convos.pm
```diff
@@ -289,37 +293,41 @@ Marcus Ramberg - C<marcus@nordaaker.com>
...
+    %= content
     <div id="vue_tooltip"><span></span></div>
     %= javascript begin
       window.Convos = {
         apiUrl:   "<%= $c->url_for('api') %>",
         indexUrl: "<%= $c->url_for('index') %>",
         wsUrl:    "<%= $c->url_for('events')->to_abs->userinfo(undef)->to_string %>",
         mode:     "<%= app->mode %>",
-        settings: <%== Mojo::JSON::encode_json(settings) %>,
+        settings: <%== Mojo::JSON::encode_json(app->config('settings')) %>,
```
jberger: this should be `to_json`

lib/Convos/Plugin.pm
jberger: It is a reasonable idea for future portability I think

lib/Convos/Plugin.pm
```diff
@@ -0,0 +1,39 @@
...
+use Mojo::Base 'Mojolicious::Plugin';
+
+use Carp 'croak';
+
+sub register { croak 'Method "register" not implemented by subclass' }
```
jberger: Doesn't the base class already croak for unimplemented `register`?

assets/vue/convos-dialog-settings.vue
jberger: This has been my hardest part about Galileo too and it isn't even front-end templates!

lib/Convos.pm
jhthorsen: Yeah, it has to be this way. After all, you still have `$app->core` available, so you can do whatever.

lib/Convos.pm
jhthorsen: To avoid double encoding..?

lib/Convos/Plugin.pm
jhthorsen: Yeah. I actually copy/pasted this line, when I still thought the plugins were supposed to be attached to Convos::Core.

assets/vue/convos-dialog-settings.vue
jhthorsen: I think I can write a AssetPack pipe, which manipulates the vue assets. The only downside with that is that you *might* need optional modules for a plugin to work.

lib/Convos/Core/Backend/File.pm
```diff
@@ -77,6 +77,19 @@ sub delete_object {
...
+  my $storage_file = $self->_settings_file($obj);
+  my $data;
+
+  -e $storage_file and eval {
+    $data = Mojo::JSON::decode_json(Mojo::Util::slurp($storage_file));
+    return next_tick $obj, $cb, '', $data;
+  };
+
+  return next_tick $obj, $cb, $@, {};
+}
```
jhthorsen: The backend can now save and load any object that has the `TO_JSON()` method.

lib/Convos/Core/Dialog.pm
```diff
@@ -36,12 +37,15 @@ sub calculate_unread {
   );
 }
 
+sub stash { Convos::Util::_stash(stash => @_) }
```
jhthorsen: I'm not sure I'm happy with this name, but I think we need something like this to prevent hard coupling between plugins and objects.

lib/Convos/Plugin/Helpers.pm
```diff
@@ -0,0 +1,87 @@
+package Convos::Plugin::Helpers;
```
jhthorsen: First core plugin! 👍

jberger:
Exactly

lib/Convos.pm
jberger: exactly

jhthorsen: Is it the same problem here? lib/Convos/Controller/Events.pm

t/config.t
```diff
@@ -1,6 +1,8 @@
 use t::Helper;
 delete $ENV{CONVOS_SECRETS};
 
+use Carp::Always;
```
jhthorsen: Going to remove this.

PR #290, opened 2016-10-31
sort dialogs without regards to leading non-word characters

jberger:
It is confusing when ##thischannel comes before #somechannel because of
the double hash. Therefore sort after stripping leading non-word
characters.


jberger:
I don't know if there is any place to test this. I have tried it locally and it works for me.

jhthorsen:
We currently don't have any frontend tests, so manual testing is sadly the only way to see if it works.

jhthorsen:
LGTM! 👍

PR #291, opened 2016-10-31
set on_connect_commands via a textarea split on newlines

jberger:
on_connect_commands is an array of commands to be run when the
connection is made. Editing an array in html isn't (super) easy, so for
a quick implementation which just uses a textarea and split the lines on
newline.

PLEASE! CHECK MY WORK HERE. I had to dig a lot and change things I wasn't 100% confident about.

### DIFF
--- a/assets/js/core/connection.js
+++ b/assets/js/core/connection.js
@@ -87,7 +87,8 @@
 
     // It is currently not possible to specify "name"
     var attrs = {
-      url: this.url
+      url: this.url,
+      on_connect_commands: this.on_connect_commands,
     };
 
     if (this.connection_id) {
--- a/assets/vue/convos-connection-editor.vue
+++ b/assets/vue/convos-connection-editor.vue
@@ -34,6 +34,9 @@
         <md-input :value.sync="username" cols="s6">Username</md-input>
         <md-input :value.sync="password" cols="s6" type="password">Password</md-input>
       </div>
+      <div class="row">
+        <md-textarea :value.sync="on_connect">On Connect Commands (one per line)</md-textarea>
+      </div>
     </template>
     <div class="row" v-if="errors.length">
       <div class="col s12"><div class="alert">{{errors[0].message}}</div></div>
@@ -68,6 +71,7 @@ module.exports = {
   data: function() {
     return {
       advanced:         false,
+      on_connect:       "",
       connection:       null,
       deleted:          false,
       errors:           [],
@@ -104,6 +108,7 @@ module.exports = {
       userinfo = userinfo.match(/[^:]/) ? userinfo + "@" : "";
       connection.user = this.user;
       connection.url = this.selectedProtocol + "://" + userinfo + this.server;
+      connection.on_connect_commands = this.on_connect.split(/\n/);
 
       if (this.nick) params.push("nick=" + this.nick);
       if (this.tls !== null) params.push("tls=" + this.tls);
@@ -128,8 +133,10 @@ module.exports = {
     },
     updateForm: function(connection) {
       var url = connection ? connection.url.parseUrl() : null;
+      var on_connect = connection ? connection.on_connect_commands.join("\n") : "";
 
       this.connection       = connection;
+      this.on_connect       = on_connect;
       this.password         = url ? url.query.password || "" : "";
       this.nick             = url ? url.query.nick || "" : "";
       this.server           = url ? url.hostPort : this.settings.default_server;
--- a/assets/vue/md-textarea.vue
+++ b/assets/vue/md-textarea.vue
@@ -1,0 +1,30 @@
+<template>
+  <div class="input-field col" :class="cols">
+    <textarea lazy v-model="value" class="validate materialize-textarea"
+      :id="id" v-el:input :name="name"
+      :readonly="readonly"
+      :disabled="disabled" :required="required"
+      @focus="hasFocus=true" @blur="hasFocus=false">
+    </textarea>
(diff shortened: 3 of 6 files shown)


assets/js/core/connection.js
```diff
@@ -87,7 +87,8 @@
 
     // It is currently not possible to specify "name"
     var attrs = {
-      url: this.url
+      url: this.url,
+      on_connect_commands: this.on_connect_commands,
```
jhthorsen: I think `on_connect_commands` should be defined inside the constructor as well, so we can watch it from Vue.

assets/vue/convos-connection-editor.vue
```diff
@@ -68,6 +71,7 @@ module.exports = {
   data: function() {
     return {
       advanced:         false,
+      on_connect:       "",
```
jhthorsen: Let's use CamelCase (onConnectCommands) in javascript, unless it maps directly to an API variable, such as "dialog_id".

assets/vue/convos-connection-editor.vue
```diff
@@ -104,6 +108,7 @@ module.exports = {
       userinfo = userinfo.match(/[^:]/) ? userinfo + "@" : "";
       connection.user = this.user;
       connection.url = this.selectedProtocol + "://" + userinfo + this.server;
+      connection.on_connect_commands = this.on_connect.split(/\n/);
```
jhthorsen: I think in some browsers, there's `\n\r` instead of just `\n`. How about doing this:
```
this.onConnectCommands.split(/\n/).map(function(cmd) {
  return cmd.trim(); // trim removes \s\r\n
});
```

assets/vue/convos-connection-editor.vue
```diff
@@ -128,8 +133,10 @@ module.exports = {
     },
     updateForm: function(connection) {
       var url = connection ? connection.url.parseUrl() : null;
+      var on_connect = connection ? connection.on_connect_commands.join("\n") : "";
```
jhthorsen: I would rather keep the variable as-is and join it when displaying it inside convos-connection-editor.vue.

assets/vue/md-textarea.vue
```diff
@@ -0,0 +1,30 @@
+<template>
+  <div class="input-field col" :class="cols">
+    <textarea lazy v-model="value" class="validate materialize-textarea"
```
jhthorsen: Why lazy?

assets/vue/md-textarea.vue
```diff
@@ -0,0 +1,30 @@
+<template>
```
jhthorsen: I like md-textarea! 👍

lib/Convos/Controller/Connection.pm
```diff
@@ -103,6 +109,9 @@ sub update {
     $state = 'reconnect' if $url->to_string ne $connection->url->to_string;
     $connection->url->parse($url);
   }
+  if (my $cmds = $json->{on_connect_commands}) {
```
jhthorsen: Pretty sure this isn't required. `$user->connection()` above will modify the original object if it exists: https://github.com/Nordaaker/convos/blob/master/lib/Convos/Util.pm#L34

lib/Convos/Controller/Connection.pm
```diff
@@ -20,7 +20,13 @@ sub create {
   }
 
   eval {
-    $connection = $user->connection({protocol => $url->scheme, name => $name});
+    $connection = $user->connection(
+      {
+        name                => $name,
+        on_connect_commands => $json->{on_connect_commands} || [],
```
jhthorsen: I think we should do some sanitising on the input. Something like this:

```
@cmds = grep {/\w/} @{$json->{on_connect_commands} || []};
```

We could even do that in the schema, but I'm not sure what is best...

assets/vue/md-textarea.vue
jberger: Honestly I didn't know what lazy is. I started by copying md-input and it must have been in there

lib/Convos/Controller/Connection.pm
jberger: Ok, I wasn't sure if there was a PATCH like behavior

lib/Convos/Controller/Connection.pm
jberger: I'm not opposed, but what is the harm of not sanitizing this?

jhthorsen: I'm afraid that the IRC server will get confused of weird commands. We could however sanities the on_connect_commands in Irc.pm before sending them to the IRC server. Maybe that makes more sense..?

assets/vue/md-textarea.vue
jhthorsen: Oh! My bad then I guess, haha

lib/Convos/Controller/Connection.pm
jhthorsen: I'm a big fan of PATCH like behaviour in general. More DWIM to me 😄

assets/vue/convos-connection-editor.vue
jhthorsen: Not in this line. Here the attribute `vue.onConnectCommands` contains a copy of the attribute `connection.on_connect_commands`. The reason for the copy is that I don't want to modify connection.on_connect_commands before you actually hit "Save"

jhthorsen:
Rebased into master (627f57a).

Thanks for the contribution!

PR #292, opened 2016-10-31
subtracting months doesn't always change month

jberger:
As documented in Time::Piece, doing month math at the end of the month
doesn't always do what you expect. As such, attempt to scroll back the
month until the month ACTUALLY rolls over.

### DIFF
--- a/lib/Convos/Core/Backend/File.pm
+++ b/lib/Convos/Core/Backend/File.pm
@@ -287,7 +287,11 @@ sub _messages {
   return [] if $args->{after} > $args->{before};
   return $args->{messages} if $args->{cursor} < $args->{after};
   my $file = $self->_log_file($obj, $args->{cursor});
-  $args->{cursor} = $args->{cursor}->add_months(-1);
+
+  # subtracting months at the end of months doesn't always work
+  my $mon = $args->{cursor}->mon;
+  $args->{cursor} = $args->{cursor}->add_months(-1)
+    while $args->{cursor}->mon == $mon;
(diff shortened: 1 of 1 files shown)


jhthorsen:
We need to test this before merging. This is a nasty bug (imo) that will probably cause regression without a test. Maybe we can extend this? https://github.com/Nordaaker/convos/blob/master/t/backend-file-messages.t#L42

pink-mist:
While I'm not quite done with it yet, and the interface may still change some, I'm making a module named ```Time::C``` which you could use quite usefully here. As an example, see the following:

```
$ perl -Ilib -MTime::C -E 'my $t = Time::C->new(2017,01,31); while ($t->year > 2015) { $t->day(0); say $t; }'
2016-12-31T00:00:00Z
2016-11-30T00:00:00Z
2016-10-31T00:00:00Z
2016-09-30T00:00:00Z
2016-08-31T00:00:00Z
2016-07-31T00:00:00Z
2016-06-30T00:00:00Z
2016-05-31T00:00:00Z
2016-04-30T00:00:00Z
2016-03-31T00:00:00Z
2016-02-29T00:00:00Z
2016-01-31T00:00:00Z
2015-12-31T00:00:00Z
```
which shows it is quite easy to get the last day of the preceding month using it =)

jhthorsen:
I managed to create a test for this, which can be seen in 82138f9.

Thanks @jberger !

jberger:
Excellent! Thanks

PR #293, opened 2016-11-06
dialog ids (channel names) may contain a dot, so use relaxed placeholder

jberger:
IRC channel names, like say #chicago.pm may contain a dot. By default
mojolicious routes strip file-extension-like endings for format
detection. To avoid this use the relaxed placeholder "#".

This of course needs a test but I don't understand how the tests know what the channel name is.


jhthorsen:
Awesome! I've added a test as well.

PR #306, opened 2017-01-06
Improve Convos Dockerfile

oliwer:
- update to Alpine 3.5
- smarter apk commands and remove apk's cache
- make the listening port configurable
  usage: docker build --build-arg port=3200
- make the convos script command configurable
  the default is still "daemon"

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,8 +5,6 @@
 FROM alpine:3.5
 MAINTAINER jhthorsen@cpan.org
 
-ARG port=3000
-
 RUN \
   apk add -U perl perl-io-socket-ssl && \
   apk add -t builddeps build-base perl-dev wget && \
@@ -33,11 +31,10 @@ ENV MOJO_REVERSE_PROXY 0
 
 # Do not change these variables unless you know what you're doing
 ENV CONVOS_HOME /data
-ENV MOJO_LISTEN http://*:$port
 ENV MOJO_MODE production
 
 VOLUME ["/data"]
-EXPOSE $port
+EXPOSE 3000
 
 CMD ["daemon"]
 ENTRYPOINT ["/convos-stable/script/convos"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,18 +2,18 @@
 #
 # BUILD: docker build --no-cache --rm -t nordaaker/convos .
 # RUN:   docker run -it --rm -p 8080:3000 -v /var/convos/data:/data nordaaker/convos
-FROM alpine:3.4
+FROM alpine:3.5
 MAINTAINER jhthorsen@cpan.org
 
-WORKDIR /
+ARG port=3000
 
-RUN apk update && \
-  apk del wget && \
-  apk add perl perl-io-socket-ssl perl-dev g++ make openssl openssl-dev wget curl && \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -2,18 +2,18 @@
 #
 # BUILD: docker build --no-cache --rm -t nordaaker/convos .
 # RUN:   docker run -it --rm -p 8080:3000 -v /var/convos/data:/data nordaaker/convos
-FROM alpine:3.4
+FROM alpine:3.5
 MAINTAINER jhthorsen@cpan.org
 
-WORKDIR /
+ARG port=3000
```
jhthorsen: What does `ARG port=3000` do?

Dockerfile
```diff
@@ -33,10 +33,11 @@ ENV MOJO_REVERSE_PROXY 0
 
 # Do not change these variables unless you know what you're doing
 ENV CONVOS_HOME /data
-ENV MOJO_LISTEN http://*:3000
+ENV MOJO_LISTEN http://*:$port
```
jhthorsen: Why does $port need to be configurable?

Dockerfile
```diff
@@ -33,10 +33,11 @@ ENV MOJO_REVERSE_PROXY 0
...
 ENV MOJO_MODE production
 
 VOLUME ["/data"]
-EXPOSE 3000
-CMD []
-ENTRYPOINT ["/convos-stable/script/convos", "daemon"]
+EXPOSE $port
+
+CMD ["daemon"]
+ENTRYPOINT ["/convos-stable/script/convos"]
```
jhthorsen: Cool! Didn't know CMD worked like this. I thought it was an alternative to ENTRYPOINT.

Dockerfile
```diff
@@ -2,18 +2,18 @@
...
-  apk del wget && \
-  apk add perl perl-io-socket-ssl perl-dev g++ make openssl openssl-dev wget curl && \
-  curl -L https://github.com/Nordaaker/convos/archive/stable.tar.gz | tar xvz && \
+RUN \
+  apk add -U perl perl-io-socket-ssl && \
+  apk add -t builddeps build-base perl-dev wget && \
+  wget -q -O - https://github.com/Nordaaker/convos/archive/stable.tar.gz | tar xvz && \
   /convos-stable/script/convos install && \
-  apk del perl-dev g++ make openssl openssl-dev curl && \
-  rm -rf /root/.cpanm /usr/local/share/man/*
```
jhthorsen: Why not delete `/usr/local/share/man/*`?

oliwer: I can confirm that the man directory is already empty. Alpine removes manpages by default.

Dockerfile
oliwer: It create a build-time variable that the user can modify when building his image.

Dockerfile
oliwer: It does not. Unless you want to run another mojo app in this container, which of course you don't.

Dockerfile
jhthorsen: Thanks for the insight!

jhthorsen:
Thanks for the contribution! 👍

PR #310, opened 2017-01-07
Add support for loading custom sass rules

oliwer:
The user can add its own rules by creating `$CONVOS_HOME/assets/sass/custom.css`

### DIFF
--- a/assets/sass/main.scss
+++ b/assets/sass/main.scss
@@ -55,3 +55,6 @@
 html {
   font-family: -apple-system, BlinkMacSystemFont, "Segoe UI", "Roboto", "Oxygen", "Ubuntu", "Cantarell", "Fira Sans", "Droid Sans", "Helvetica Neue", sans-serif;
 }
+
+// Add your own rules in $CONVOS_HOME/assets/sass/custom.css
+@import "http://local/custom/asset/sass/custom.scss";;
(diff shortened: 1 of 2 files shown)


jhthorsen:
Could you also extend https://github.com/Nordaaker/convos/blob/master/t/web-custom-assets.t ?

Just add something like this:

```
my $custom = File::Spec->catfile($ENV{CONVOS_HOME}, qw(assets sass _custom.scss));
# open and write to $custom, and then test that the custom styles are included with something like this:
->content_like(qr{.convos-message.same-user .title});
```

assets/sass/main.scss
```diff
@@ -55,3 +55,6 @@
 html {
   font-family: -apple-system, BlinkMacSystemFont, "Segoe UI", "Roboto", "Oxygen", "Ubuntu", "Cantarell", "Fira Sans", "Droid Sans", "Helvetica Neue", sans-serif;
 }
+
+// Add your own rules in $CONVOS_HOME/assets/sass/custom.css
+@import "http://local/custom/asset/sass/custom.scss";;
```
jhthorsen: Could you also extend https://github.com/Nordaaker/convos/blob/master/t/web-custom-assets.t ?

Just add something like this:

```
my $custom = File::Spec->catfile($ENV{CONVOS_HOME}, qw(assets sass _custom.scss));
# open and write to $custom, and then test that the custom styles are included with something like this:
->content_like(qr{.convos-message.same-user .title});
```

jhthorsen:
Perfect!

PR #316, opened 2017-01-15
Derive connection name based on ZNC usernames

tyldum:
ZNC is an IRC bouncer that supports multiple networks.
    By detecting the full login signature of <user>@<useragent>/<network> we
    can utilize this information.
    Even though <useragent> is optional in ZNC, it is mandatory for Convos
    to be able to securely detect this style login.
    
    The username given in Convos would look something like this:
        myuser@Convos/freenode
    And would produce a connection name of "freenode" wihin Convos, which
    would be displayed as "irc-freenode" in the frontend.

The bigger issue is a concept of networks within Convos, but this is good enough for me.


jhthorsen:
Looks great!

PR #323, opened 2017-03-09
Recover password

jhthorsen:
Creating a pull request, because I'm unsure about the security implications. I will add comments to the parts in the code.


jhthorsen:
I've changed how session secrets are generated: 211d16b

PR #329, opened 2017-04-03
Feature: pastebin

jhthorsen:
This PR implements a pastebin internally to Convos. It is implemented in the backend, meaning that a given connection class can detect a long message and dispatch it to the "paste-service". This is done automatically, without asking the user.

The nice thing is that it's implemented as a plugin, so if someone wants to make a "gist" backend, then it shouldn't be too hard.

This is how you enter a paste:

screen shot 2017-04-04 at 09 19 33

This is how it looks to a Convos user: (The paste is embedded using LinkEmbedder.pm)

170404-convos-pastebin-linkembedder

This is how it looks if you open the link in a browser: (The link is available from the public)

Note also that the paste support syntax highlighting, using https://highlightjs.org/

170404-convos-pastebin-public


t/web-x-request-base.t
```diff
@@ -8,4 +8,6 @@ $t->get_ok('/x/request/base')->status_is(200)->content_is('/foo');
 $t->get_ok('/x/request/base', {'X-Request-Base' => 'http://example.com/sub'})->status_is(200)
   ->content_is('/sub/foo');
 
+like $t->app->core->base_url, qr{^http://example\.com}, 'core base_url';
```
marcusramberg: should check that path is /sub for base_url

lib/Convos/Core/Connection/Irc.pm
```diff
@@ -375,7 +375,16 @@ sub _send {
...
 
   if (MAX_BULK_MESSAGE_SIZE < @messages) {
-    $self->user->core->backend->emit_single(multiline_message => $self, $message, $cb);
+    $self->user->core->backend->emit_single(
+      multiline_message => $self,
+      \$message,
+      sub {
+        my ($backend, $err, $file) = @_;
+        return $self->$cb($err) if $err;
+        my $url = $self->user->core->base_url->clone->path($file->uri);
```
marcusramberg: Needs to use base_url as base, not basis for created url.

lib/Convos/Plugin/Paste.pm
```diff
@@ -0,0 +1,100 @@
...
+      my ($delay, $err, $data) = @_;
+      die $err if $err;
+      $data->{created_at} = Mojo::Date->new($data->{created_at})->to_datetime;
+      $c->render(paste => file => $data, render_js => 0);
+    }
+  );
+}
+
+package    # \
+  Convos::Plugin::Paste::File;
```
marcusramberg: This is ugly :-1:

marcusramberg: I suggest making this a public class.

PR #332, opened 2017-04-13
Fixing memory leaks

jhthorsen:
It seems like Convos is leaking memory, but I cannot seem to figure out where it happens.

The original code had `weaken()` thrown all over the place, without any tests. I want to try to be a bit smarter this time and try to pin point where the problem might be.

### DIFF
--- a/lib/Convos/Controller/Url.pm
+++ b/lib/Convos/Controller/Url.pm
@@ -11,7 +11,7 @@ sub info {
   }
   if (my $link = $self->app->_link_cache->get($url)) {
     $self->res->headers->header('X-Cached' => 1);    # for testing
-    return $self->respond_to(json => {json => $link}, any => {text => $link->html});
+    return $self->respond_to(json => {json => $link}, any => {text => $link->{html} // ''});
   }
 
   $self->delay(
@@ -23,7 +23,7 @@ sub info {
         $self->stash(status => $link->error->{code} || 500);
       }
       else {
-        $self->app->_link_cache->set($url => $link);
+        $self->app->_link_cache->set($url => $link->TO_JSON);
         $self->res->headers->cache_control('max-age=600');
       }
 
--- a/lib/Convos/Core/Backend.pm
+++ b/lib/Convos/Core/Backend.pm
@@ -58,14 +58,12 @@ sub users {
 sub _setup {
   my $self = shift;
 
-  Scalar::Util::weaken($self);
   $self->on(
     connection => sub {
       my ($self, $connection) = @_;
       my $cid = $connection->id;
       my $uid = $connection->user->id;
 
-      Scalar::Util::weaken($self);
       $connection->on(
         message => sub {
           my ($connection, $target, $msg) = @_;
--- a/lib/Convos/Core/Connection/Irc.pm
+++ b/lib/Convos/Core/Connection/Irc.pm
@@ -30,7 +30,6 @@ has _irc => sub {
   my $self = shift;
   my $irc = Mojo::IRC::UA->new(debug_key => join ':', $self->user->email, $self->name);
 
-  Scalar::Util::weaken($self);
   $irc->name("Convos v$Convos::VERSION");
   $irc->parser(Parse::IRC->new(ctcp => 1));
   $irc->unsubscribe('message');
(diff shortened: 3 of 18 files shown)


lib/Convos/Controller/Url.pm
```diff
@@ -23,7 +23,7 @@ sub info {
         $self->stash(status => $link->error->{code} || 500);
       }
       else {
-        $self->app->_link_cache->set($url => $link);
+        $self->app->_link_cache->set($url => $link->TO_JSON);
```
jhthorsen: This avoids the cycle below, which I don't think is a "real" cycle.

```
# Cycle #1
#     Convos A->{_link_cache} => Mojo::Cache B
#     Mojo::Cache B->{cache} => %C
#     %C->{http://127.0.0.1:63544} => LinkEmbedder::Link::Basic D
#     LinkEmbedder::Link::Basic D->{ua} => Mojo::UserAgent E
#     Mojo::UserAgent E->{connections} => %F
#     %F->{8888b3a629283ac9742be4004f920fde} => %G
#     %G->{cb} => &H
#     closure &H, $self => $I
#     $I => LinkEmbedder::Link::Basic D
```

lib/Convos/Core/Backend.pm
```diff
@@ -58,14 +58,12 @@ sub users {
...
   my $self = shift;
 
-  Scalar::Util::weaken($self);
   $self->on(
     connection => sub {
       my ($self, $connection) = @_;
       my $cid = $connection->id;
       my $uid = $connection->user->id;
 
-      Scalar::Util::weaken($self);
```
jhthorsen: I don't understand why `weaken()` is not required here and/or above. (Same change is in File.pm)

lib/Convos/Core/Connection/Irc.pm
```diff
@@ -435,8 +432,7 @@ sub _steal_nick {
   my $self = shift;
   my $tid;
 
-  Scalar::Util::weaken($self);
-  $tid = $self->_irc->ioloop->recurring(
```
jhthorsen: I don't get this either: If I avoid using the `ioloop` from Mojo::IRC, then I also avoid a memory cycle..?

Also, why is `weaken()` not required above?

lib/Convos/Core/User.pm
```diff
@@ -33,6 +32,7 @@ has_many connections => 'Convos::Core::Connection' => sub {
 
   my $connection = $class->new($attrs);
   warn "[@{[$self->email]}] Emit connection for id=@{[$connection->id]}\n" if DEBUG;
+  Scalar::Util::weaken($connection->{user} = $self);
```
jhthorsen: This actually fixes a bug, but I doubt this is where the problem is.

jhthorsen:
I've started checking if 0.99_30 still have a memory issue, but it might take a day before I have some good data. I'm currently running `while sleep 900; do date;ps axfu --cols 200|grep 8081|grep -v grep|awk '{print $4 "/" $5 "/" $6}';done >> ~/convos.mem` on the "dogfood" server. Convos is currently using 3.1% memory, but before I upgraded from 0.99_29 earlier today, it used 9% memory.

```
$ tail -f convos.mem
Thu Apr 13 09:17:16 CDT 2017
3.1/147840/64140
Thu Apr 13 09:32:16 CDT 2017
3.1/147840/64140
Thu Apr 13 09:47:16 CDT 2017
3.1/147840/64140
Thu Apr 13 10:02:16 CDT 2017
3.1/150460/64464
Thu Apr 13 10:17:16 CDT 2017
3.1/150460/64464
Thu Apr 13 10:32:16 CDT 2017
3.1/150460/64464
Thu Apr 13 10:47:16 CDT 2017
3.1/150460/64464
Thu Apr 13 11:02:16 CDT 2017
3.1/150460/64464
Thu Apr 13 11:17:16 CDT 2017
3.1/150460/64464
Thu Apr 13 11:32:16 CDT 2017
3.1/150460/64464
Thu Apr 13 11:47:16 CDT 2017
3.1/151096/65060
```

jhthorsen:
The failing travis tests can be fixed with something like this:

```
sub Mojo::IRC::ioloop { return Mojo::IOLoop->singleton }
```

lib/Convos/Core/Backend.pm
marcusramberg: Looks like $self is not used at all and thus not captured?

jhthorsen: `$self->emit(..)` is called in both of the event handlers.

jhthorsen:
Closing this because of inactivity.

PR #347, opened 2017-05-03
Add snap install documentation

flexiondotorg:
I see you've published Convos in the snap store! Thanks for snapping Convos it's pretty fantastic :-) I noticed you're missing snap install instructions meaning that many Linux users won't realise they could be installing an always updated snap.

This pull request adds installation instructions for users of [snap enabled Linux distributions](https://snapcraft.io/docs/core/install). Snaps enable ISVs and projects to make their software available to millions of Linux users via the snap store and directly control delivery of software updates to their users. Your software can be installed with `snap install convos` simplifying the installation instructions for your users.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,13 +13,27 @@ frontend is held together by the progressive JavaScript framework
 
 # Quick start guide
 
+## Snap Install
+
+Install Convos in seconds on [Ubuntu and other snap supported Linux distributions](https://snapcraft.io/docs/core/install) with:
+
+    snap install convos
+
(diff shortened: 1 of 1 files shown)


jhthorsen:
Thanks for contributing @flexiondotorg! I wonder if this should also be mentioned in https://github.com/Nordaaker/convos/blob/gh-pages/doc/getting-started.md

marcusramberg:
Will give it a try tomorrow. Good exercise. Going to a Russian micro brewery tonight.  :)

PR #357, opened 2017-07-25
As a user, I want to clear the notifications count when I show the notifications tab

marcusramberg:
This pull request implements an option that makes convos automatically clear notifications when you open the notifications panel. Decided to leave the manual clear button in place to allow clearing notifications that come in while the panel is already open.


jhthorsen:
The code looks correct, but I haven't tested it.

Maybe you want to write a test as well? Not sure if this is good enough to get inspiration from: https://github.com/Nordaaker/convos/blob/master/t/selenium-settings.t

jhthorsen:
I'm gonna close this, since @marcusramberg has implemented this in the new Svelte based version as well.

PR #369, opened 2018-04-27
Add service worker and cache asssets WIP: Implemnt Service Workers

marcusramberg:
This PR implements service workers, part of PWA effort for Convos


lib/Convos.pm
```diff
@@ -69,6 +69,7 @@ sub startup {
   # Add basic routes
   $r->get('/')->to(template => 'index')->name('index');
   $r->get('/err/500')->to(cb => sub { die 'Test 500 page' });
+  $r->get('/sw' =>  [ format => 'js']);
```
jhthorsen: Could we name it "service-worker" instead of "sw"? "sw" for me could be star-wars.js 😆

lib/Convos.pm
```diff
@@ -138,14 +139,39 @@ sub _assets {
...
-      picture_aspect => "shadow",
-      manifest       => {name => "Convos", display => "standalone", orientation => "portrait"},
-      theme_color    => "#00451D"
+      picture_aspect => 'shadow',
+      manifest       => {name => 'Convos', display => 'standalone', orientation => 'portrait'},
+      theme_color    => '#00451D'
     },
+    firefox_app => {
+        picture_aspect=> 'circle',
+        keep_picture_in_circle=> 'true',
```
jhthorsen: Looks like this part isn't tidied.

Another thing not related to Convos: Should there be a config file that AssetPack::Pipe::Favicon looks up, so we don't need all of this inside the application?

templates/layouts/convos.html.ep
```diff
@@ -1,6 +1,6 @@
 % my $description = "A chat application that runs in your web browser";
 <!DOCTYPE html>
-<html data-framework="vue">
+<html data-framework="vue" lang="en">
```
jhthorsen: What's the benefit of lang=en?

templates/layouts/convos.html.ep
```diff
@@ -26,5 +26,17 @@
     %= content
     <div id="vue_tooltip"><span></span></div>
     %= asset 'convos.js' if 200 == (stash('status') // 200)
+    <script>
```
jhthorsen: Could we move this to assets/js/main.js?

PR #383, opened 2019-05-30
window.document.hasFocus instead of window.hasFocus?

Wikiwide:
It could potentially be a setting. But I think that it's better to show notification, if possible. Because user may be at a different tab in the same window?..


marcusramberg:
Seems sane.

PR #387, opened 2019-09-14
Doc update and minor fixes

jjatria:
While reading the documentation I found a number of small nits that I thought I might as well fix.

One of them was in the LOC stats table, which I realised hadn't been updated in a couple of years, so I thought I might as well.

I'm really liking Convos, by the way. :smiley:


jhthorsen:
Thanks! Looks great :

PR #395, opened 2019-10-28
Feature/jsdoc

jhthorsen:
This PR holds both a jsdoc "theme", so it looks like the documentation is part of Convos, but it also holds the generated documentation. I was thinking it might be useful to have bundled, but is it really? Maybe I should just include the /docs/ endpoint and let the developer know how to generate the docs instead?

What do you think @marcusramberg?

Also, the actual documentation was pushed directly to "master": 51431a180440e208b5f53a42dc7546f086a4669f


jhthorsen:
Nevermind. I won't bundle the generated docs, since they will be outdated.

PR #400, opened 2019-10-30
Add handling for connection protocols, closes #311

### LINKED ISSUE #311: Should be able to handle "irc://..." links
Not quite sure how to register links in the browser, but at least "irc://" links from within Convos should result in a new connection and a new dialog.

The first link to handle is the link in the text in convos-resources.vue:

> Wanna talk to us? Join the "#convos" channel on irc.freenode.net

jhthorsen:
This PR adds the following features:

- Will ask the user to register the "irc" protocol, so if they click on a "irc://" link it will open in Convos
- The irc://... link will be passed on to /add/connection?uri=...
- &lt;ConnectionForm/> will parse the uri and see if:
  - The connection with the same hostname and the conversation exists. If so, the user is taken to /chat/irc-whatever/dialogName.
  - Check if just the connection exists. If so, the user is taken to /add/conversation?connection_id=irc-whatever&dialog_id=dialogName
  - If the connection does not exist, the &lt;ConnectionForm/> will be populated.

Other features/fixes:

- <ConnectionForm/> got a "dialog_id" field, so the user can add a single dialog to join after connected.
- ConnUrl.pathname accessor was added
- md() will convert anything that starts with a scheme into a link, and not just "http" and "https".

### DIFF
--- a/assets/components/ConnectionForm.svelte
+++ b/assets/components/ConnectionForm.svelte
@@ -45,7 +45,28 @@ function defaultsToForm() {
   formEl.nick.value = user.email.replace(/@.*/, '').replace(/\W/g, '_');
   useTls = true;
   wantToBeConnected = true;
-  urlToForm(formEl);
+
+  const url = new URL(location.href);
+  if (!url.searchParams.get('uri')) return urlToForm(formEl, url);
+
+  const connURL = new ConnURL(url.searchParams.get('uri'));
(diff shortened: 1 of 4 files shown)


jhthorsen:
I want to handle this on the server side instead.

jhthorsen:
Replaced by c128de7f65420b4336836eb048b095590138be6f

PR #405, opened 2019-11-03
Implement nord color scheme

marcusramberg:
Based on https://www.nordtheme.com

### DIFF
--- a/assets/sass/_themes.scss
+++ b/assets/sass/_themes.scss
@@ -8,6 +8,7 @@ $sidebar-item-icon-width: 1.5rem;
 
 @import '../../node_modules/highlight.js/scss/atom-one-light.scss';
 
+@import 'themes/nord';
 @import 'themes/dark';
 @import 'themes/light';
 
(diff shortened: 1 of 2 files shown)


assets/sass/_themes.scss
```diff
@@ -10,8 +10,9 @@ $sidebar-item-icon-width: 1.5rem;
 
 @import 'themes/dark';
 @import 'themes/light';
+@import 'themes/nord';
```
jhthorsen: There's a missing part  in my [post](https://convos.by/2019/11/2/custom-styling.html) about the order: When it's a custom theme, the theme should be before "themes/light". Please re-order it.

marcusramberg:
EId4B18WsAM79Tv
Will fix reordering and get rid of the asset changes.

jhthorsen:
Looks good! Could this be the default dark mode?

marcusramberg:
I think I would prefer for the green one to remain the default, for branding, but I prefer this one myself :)

jhthorsen:
Maybe it’s time to rebrand 😉

PR #409, opened 2019-11-07
Fix the color on "textarea" focused element on dark theme

serhack:
**BEFORE**
image

**AFTER**
image


jhthorsen:
Thanks for the fix, but unfortunately it wasn't correct because of a "bug" (shortcut?) I made in the CSS rules. This have now been fixed in 33f2117, where I added a new variable `--input-focus-placeholder-color`.

PR #418, opened 2019-11-25
Add LDAP support to Convos, closes #65

### LINKED ISSUE #65: Add LDAP support
It would be nice to have some other authentication mechanisms and LDAP in particular. I'm sure I'm not the only one keen on running something like this on a subnet far far away with sufficient infrastructure for user authentication.

jhthorsen:
Several people have asked for LDAP support, so I thought it might be about time to add it.

I have no idea how LDAP works though, so I need help setting up a test environment so I can see that it actually works. I've so far managed to install openldap, but I'm unable to add a test-user.

I would also like feedback on the default values.

### DIFF
--- a/lib/Convos/Plugin/Auth/LDAP.pm
+++ b/lib/Convos/Plugin/Auth/LDAP.pm
@@ -1,0 +1,143 @@
+package Convos::Plugin::Auth::LDAP;
+use Mojo::Base 'Convos::Plugin::Auth';
+
+use Convos::Util qw(DEBUG require_module);
+use Mojo::Promise;
+
+has _ldap_options => undef;
+has _ldap_url     => undef;
+has _reactor      => sub { Mojo::IOLoop->singleton->reactor };
+
+sub register {
+  my ($self, $app, $config) = @_;
+
+  # Allow ldap url with options: ldaps://ldap.example.com?debug=1&timeout=10
+  my $ldap_url = Mojo::URL->new($ENV{CONVOS_AUTH_LDAP_URL} || 'ldap://localhost:389');
+  $self->_ldap_options($ldap_url->query->to_hash);
+  $self->_ldap_options->{timeout} ||= 10;
+  $self->_ldap_url($ldap_url->query(Mojo::Parameters->new));
+
+  # Make sure Net::LDAP is installed
+  require_module('Net::LDAP');
+
+  $app->helper('auth.login_p' => sub { $self->_login_p(@_) });
+  $app->log->debug("Loaded Convos::Plugin::Auth $ldap_url");
+}
+
+sub _bind_params {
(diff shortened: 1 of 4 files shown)


tboerger:
For testing purpose you could use https://github.com/butonic/devldap pretty easily, you could even write integration tests with that. Just feed it with an JSON file. An alternative to that could also be https://github.com/glauth/glauth which got support for a TOML file as a user datasource.

jhthorsen:
I would greatly appreciate some help to finish this branch @tboerger. Can you help?

tboerger:
I'm not into Perl, but I could test and provide you with debug logs :)

tboerger:
As far as I understand the Perl code I would expect a different flow.

### Config

* ldap url
* bind dn 
* bind password 
* user/login filter 

### Flow

* ldap connect: connect with the ldap url
* ldap bind: bind the request to dn and password 
* ldap search: search the user based on the filter, replace a username placeholder with the requesting username 
* check password: user the cn from ldap search and check the password with another ldap bind request 
* login error/success 

### Example 

The user filter should be in a form like this:

(&(objectclass=person)(|(uid=USERNAME_PLACEHOLDER)(mail=USERNAME_PLACEHOLDER)))

jhthorsen:
Thanks for the tip about devldap! Once I figured out how to compile it, it was super easy to use! 😄 

```
$ CONVOS_AUTH_LDAP_URL=ldap://127.0.0.1:10389?debug=1 ./script/convos exec prove -vl t/web-ldap.t
...
All tests successful.
Files=1, Tests=16,  1 wallclock secs ( 0.03 usr  0.01 sys +  0.64 cusr  0.08 csys =  0.76 CPU)
```

I'm going to merge this branch now, but unfortunately I have a blocking bug in the frontend code, so I cannot make a release just yet. You can however checkout "master" and try the new version out, since "master" does not include the frontend bug.

jhthorsen:
You can read about how to use the LDAP plugin here: https://github.com/Nordaaker/convos/blob/master/lib/Convos/Plugin/Auth/LDAP.pm#L104

Once a new release is made, I will update the website as well to make it easier to read. Please have a look at https://github.com/Nordaaker/convos/blob/master/lib/Convos/Plugin/Auth/LDAP.pm#L118 to see if you think the defaults make sense.

jhthorsen:
I don't really know how LDAP works, but I'm guessing if you want the above search to work you need to start convos like this:

```
CONVOS_AUTH_LDAP_DN="(&(objectclass=person)(|(uid=%uid)(mail=%email)))" \
CONVOS_AUTH_LDAP_URL="ldap://localhost:389" \
CONVOS_PLUGINS=Convos::Plugin::Auth::LDAP \
./script/convos daemon
```

PR #440, opened 2020-02-09
Add to Docker section about SELinux and exposing ports

losuler:
I found these two notes were helpful when I was setting up Convos with Docker. 

The first one about exposing to localhost only (1c913f2) I thought was useful as I was only using it locally and didn't want to expose it anywhere else unnecessarily or accidentally.

The second one about Docker volumes and SELinux (18cccf0) is important because without it nothing can be written to the volume on the host (`$HOME/convos/data`).

### DIFF
--- a/doc/getting-started.md
+++ b/doc/getting-started.md
@@ -104,6 +104,12 @@ docker run -it -p 8080:3000 \
 The last command will make Convos available on http://localhost:8080, and
 persist data in `$HOME/convos/data`.
 
+For Linux distributions with SELinux Enforcing policy (e.g. CentOS, Fedora or RHEL) append `:z` to volumes:
+
+```bash
+-v $HOME/convos/data:/data:z
+```
+
 There are some [alternative tags](https://hub.docker.com/r/nordaaker/convos/tags)
 available, but we suggest using the "stable" release.
 


doc/getting-started.md
```diff
@@ -104,6 +104,18 @@ docker run -it -p 8080:3000 \
 The last command will make Convos available on http://localhost:8080, and
 persist data in `$HOME/convos/data`.
 
+To expose Convos to localhost only, prepend `127.0.0.1` like so:
+
+```bash
+-p 127.0.0.1:8080:3000
```
jhthorsen: I think this is a bit too much information here. I'm not entirely sure how frequent people start Convos on localhost only.

marcusramberg: I tend to agree with jhthorsen about this, and it's not really convos specific info.

doc/getting-started.md
```diff
@@ -104,6 +104,18 @@ docker run -it -p 8080:3000 \
...
+```
+
+For Linux distributions with SELinux Enforcing policy (e.g. CentOS, Fedora or RHEL) append `:z` to volumes:
+
+```bash
+-v $HOME/convos/data:/data:z
+```
+
```
marcusramberg: I think this is a useful tip tho, and would like to keep this part.

doc/getting-started.md
losuler: I felt like this might of been the case, I did feel the same way a bit. I've simply removed the commit.

PR #448, opened 2020-02-29
fix exception when searching too long a period

preaction:
This fixes an exception I found in my logs about `Can't locate object
method "resolve" via package "Mojo::IOLoop" at
/app/script/../lib/Convos/Core/Backend/File.pm line 122.` Not sure
what's sending those requests, but it seems like the intent was to just
return nothing.

### DIFF
--- a/lib/Convos/Core/Backend/File.pm
+++ b/lib/Convos/Core/Backend/File.pm
@@ -118,7 +118,7 @@ sub messages_p {
 
   # Do not search if the difference between "before" and "after" is more than 12 months
   # This limits the amount of time that could be spent searching and it also prevents DoS attacks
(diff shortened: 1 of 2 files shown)


jhthorsen:
Yikes. That’s pretty bad :/ I am curious what’s sending those requests though...

preaction:
Yeah, for some reason my browser is not letting me log in to Convos, and when I try logging in I get that exception as an error... The request being sent is `GET http://irc.preaction.me/api/user.json?connections=true&dialogs=true`, but I haven't had a chance to look in to where that gets routed and how it makes it to that exception...

preaction:
Running the latest `master` branch, I no longer get this error and am able to log in as expected. Thanks!

PR #465, opened 2020-05-14
Do not use "convos green" for button and links but prefer red.

thibaultduponchelle:
Hello,

I really like (new ?) high-contrast theme especially light version but I'm not very happy with keeping "convos green" for links and buttons.

I propose to change it to red for both light and dark version with minor diffs (disabled button is darker on dark version where it is lighter on light version + link color not exactly the same)

It could be also a new theme if some people do not agree with my vision and want to keep the current themes.

Or you can edit it or just get inspired for doing something else and decline this PR, I'm not an UX designer and it's probably not perfect :)

Here is a preview : https://imgur.com/a/pFRwn7x

Best regards.

Thibault

### DIFF
--- a/public/themes/high-contrast_color-scheme-dark.css
+++ b/public/themes/high-contrast_color-scheme-dark.css
@@ -6,17 +6,19 @@
 :root {
   --body-bg: #222222;
   --text-color: #fdfdfd;
-  --link-color: #b4eca0;
+  --link-color: #ff4444;
   --hr-border: 1px solid #333333;
   --error-color: #ee2222;
   --success-color: #338833;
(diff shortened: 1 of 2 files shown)


marcusramberg:
Making it as a separate theme seems more logical to me 🤷

thibaultduponchelle:
> Making it as a separate theme seems more logical to me shrug

Ok then #466 :smile:

jhthorsen:
I don't think this needs to be a new theme, since it is pretty much what I had when I started out making that high-contrast theme. I'm going to take this PR 👍 

I really appreciate the preview as well, but could you embed it into the PR next time?

PR #466, opened 2020-05-14
New theme blood

thibaultduponchelle:
Hello,

I really like (new ?) high-contrast theme especially light version but I'm not very happy with keeping "convos green" for links and buttons.

I propose this new theme with links and buttons red and a very light gray on right sidebar.
It is based on light color scheme (I do not added a the dark version, because the light version seems better to me)

This PR supersede the #465 following marcus comment :)

Best regards.

Thibault

### DIFF
--- a/public/themes/blood.css
+++ b/public/themes/blood.css
@@ -1,0 +1,34 @@
+/*
+ * name: Blood
+ * color-scheme: light
+ */
+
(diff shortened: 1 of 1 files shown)


jhthorsen:
I decided to take #465 instead.

PR #467, opened 2020-05-15
Better button outline + light gray for right sidebar (ligth version)

thibaultduponchelle:
Hello,

Just a small update to retrofit the button focus outline and the very light gray on the right sidebar (only light version)

Preview : 

Focus outline before change : 
convos13

Focus outline after : 
convos14
(maybe not perfect as is but at least should not be green)

Sidebar after change : 
convos12

Best regards.

Thibault

PR #468, opened 2020-05-15
Why convos does not reload after a code update?

thibaultduponchelle:
Hello,

FAQ addition as suggested in PR #467 

> Please note that when I try to update my instance to this code (or even my previous merged pr) + master code it does not start...

Note: this was fixed with a restart of Convos. Maybe this could go into the faq?

_Originally posted by @jhthorsen in https://github.com/Nordaaker/convos/pull/467#issuecomment-629081457_

Best regards.

Thibault


doc/faq.md
```diff
@@ -176,6 +176,29 @@ re-install Convos:
 
 Please [contact us](/doc/#get-in-touch) if the above instructions do not work.
 
+## Why convos does not restart after an upgrade of convos code?
```
jhthorsen, line 179: I think this heading sounds disconnected with the rest of the text.

doc/faq.md
```diff
@@ -176,6 +176,29 @@ re-install Convos:
...
+
+Some changes in the convos code do not require any special operation (for instance if you modify only assets and/or themes).
```
jhthorsen, line 181: Could this be more specific? "Like changes in the `public/` directory does not require you to restart Convos"?

doc/faq.md
```diff
@@ -176,6 +176,29 @@ re-install Convos:
 
 Please [contact us](/doc/#get-in-touch) if the above instructions do not work.
 
+## Why convos does not restart after an upgrade of convos code?
+
+Some changes in the convos code do not require any special operation (for instance if you modify only assets and/or themes). 
+
+If you update the convos code and get stuck to the loading page after a refresh like this :
```
jhthorsen, line 183: I'm not a fan of space in front of colon. (This occurs multiple times in this PR)

doc/faq.md
```diff
@@ -176,6 +176,29 @@ re-install Convos:
...
+
+<img height="300px" src="../public/screenshots/2020-05-15-09-40-05-reload-error.png" alt="Stuck during reload">
+
+Then you likely have pulled changes that require you to do a service restart. 
+
+Then simply execute :
+
+```
+$ killall convos
```
jhthorsen, line 192: Maybe this needs an alternative like `kill $PID_OF_RUNNING_CONVOS`?

doc/faq.md
thibaultduponchelle: I tried to be more specific without being exhaustive (because I don't know what is the exact list of things that we can change without having to restart convos services). Is it better for you?

doc/faq.md
thibaultduponchelle: I think I fixed this

doc/faq.md
thibaultduponchelle: I added this alternative, is it ok for you ?

doc/faq.md
thibaultduponchelle: No solution for this...

doc/faq.md
jhthorsen: Oh! I was wrong: The compiled assets (css+js) will be in public/asset/, and requires you to restart Convos. Maybe it would be better to just say that after you update (git pull, or whatever), you need to restart Convos. If you want to modify themes while Convos is running, then modify them in `$CONVOS_HOME/themes`.

jhthorsen: Again, this is why I think it might be better with an upgrade guide.

thibaultduponchelle: Ok, what if we just close this PR ? 

There are some problems that are not easy to fix without a major rework.

It could just serve as an inspiration guide when someone will write this updgrade guide :smile:

thibaultduponchelle:
Hello,

Will close this PR, it's not exactly what we want, a bigger "How to restart convos" doc would be better than this.

Feel free to reuse portions of this if needed :)

Thibault

PR #469, opened 2020-05-17
New theme "South" 

thibaultduponchelle:
Hello,

I propose this new theme based on blue/white/mustard.

south1

south2

south3

south4

south5

south6

south7

south8

south9

south10

Best regards.

Thibault

### DIFF
--- a/public/themes/south.css
+++ b/public/themes/south.css
@@ -4,7 +4,7 @@
 
 :root {
   --body-bg: #fdfdfd;
-  --link-color: #ba8c58;
+  --link-color: #993300;
   --hr-border: 1px solid #ba8c58;
   --error-color: #ee2222;
   --success-color: #338833;
--- a/public/themes/south.css
+++ b/public/themes/south.css
@@ -1,0 +1,32 @@
+/*
+ * name: South
+ */
+
+:root {
+  --body-bg: #fdfdfd;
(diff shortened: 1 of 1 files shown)


jhthorsen:
Thanks for the contribution, but I'm not completely sold on the brown on top of the blue in the sidebar. I think the blue "screams" a bit too much. Also, could you increase the contrast for brown links?

Note to myself: I think I'll revert 1088c65b02668af01f5680b2d7a86f6a22db5a31 (#439), since the background is hard to use with themes and the photo does not look too good on big screens.

thibaultduponchelle:
Hmm.. I'm very happy with this theme and don't want to change the colors.

I knew it was a bit eccentric but wanted to have an opinion to see if it worth to share :smile: 

I will close this PR then and keep it for me :)

Maybe a cool convos feature could be to add a "alternative repo" input in convos settings to link an url that contains non-official themes (but I can also simply keep it in my convos instance for the moment)

thibaultduponchelle:
Increased contrast for links.

Actually using a color between mustard and red, looks fine and stronger :)

red-mustard1

red-mustard2

red-mustard3

jhthorsen:
I'll take this theme because I want users to be able to select from more themes.

Thank you for fixing the contrast! 👍

PR #476, opened 2020-05-24
New color scheme for theme "Nord".

thibaultduponchelle:
Hello,

This PR actually contains 2 things. 

A small correction for disabled button color in theme "Nord" : 

Before : 
norddisabledbefore

After : 
norddisabledafter

And a new variation of the "Nord" color theme. It's more a "reversed Nord".

The idea is simple, exchange bg colors from panels to have dark on left panel and light on middle/right : 

nordvsdron

Some details or other pages : 

dron0

dron1

dron2

dron3

dron4


marcusramberg:
I like the theme, but I'd prefer another name, like Nord-alt or Nord-light, The colors are still the ones from nordtheme.com/docs/colors-and-palettes/ even if the order is changed.

jhthorsen:
@marcusramberg: The idea here is that the addition of this theme variant, then "Color scheme" "Auto" will switch between the light and dark version of "Nord". You can however force it to "Dark" if you want. I think naming this "Nord-light" will make the "Color scheme" selector even more confusing, ref #477.

marcusramberg:
I was referring to the filename from the diff, dron.css

jhthorsen:
Ah! Good catch @marcusramberg :) @thibaultduponchelle: Call it `nord_color-scheme-light.css`, but don't rename the origin `nord.css` since it *might* cause some issues.

public/themes/dron.css
```diff
@@ -0,0 +1,95 @@
+/*
```
jhthorsen, line 1: This file needs to be renamed.

thibaultduponchelle: Good point, done

Issue #477, opened 2020-05-24
Color schemes names somewhat hardcoded


thibaultduponchelle:
**Describe the bug**
We can use only "light" and "dark" as color schemes names

**To Reproduce**
1. Go to 'public/themes'
2. Edit 2 themes having a color-scheme
3. Refresh your convos page
4. See discrepancy in settings/account

**Screenshots**
![theme-color-scheme](https://user-images.githubusercontent.com/580360/82758002-3e353480-9de4-11ea-82b6-5f9b34bccee4.png)

Maybe it's a limitation (but it is not mentioned in https://convos.by/2020/5/14/theming-support-in-4-point-oh.html and the fact it appears well computed from css file header between parenthesis suggests the contrary)

thibaultduponchelle:
And, while I opened an issue about it, why do the input is there when the theme has no color scheme ? 

![color-theme](https://user-images.githubusercontent.com/580360/82758239-a0426980-9de5-11ea-8ccb-deb466ffbf74.png)

jhthorsen:
@thibaultduponchelle: I guess I should mention that explicitly in theming-support-in-4-point-oh.html, but the idea is that only "dark", "light" and "auto" is supported, since that is what [prefers-color-scheme](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme) supports.

The reason why I solved the "Color scheme" with a dropdown is to limit the number of choices in the "Theme" dropdown. I do however think that the "Color scheme" dropdown should probably get disabled if the selected "Theme" only supports "default" and no other variants.

Thoughts?

marcusramberg:
I'm pro disabling them when there's no option to be made.

jhthorsen:
I'm closing this since #478 and #482 is merged.

Thank you @thibaultduponchelle 😄

PR #478, opened 2020-05-25
Document the naming of color schemes in the blog post

thibaultduponchelle:
I propose something like this for #477 

This is where I looked for this info.

Regards.


jhthorsen:
Thanks 👍

PR #479, opened 2020-05-25
There is ALWAYS an "hacker" skin

thibaultduponchelle:
Screenshot from 2020-05-25 19-31-42

settings

hackerr

This is the theme produced along this [walkthrough](https://github.com/thibaultduponchelle/convos-theme-walkthrough/blob/master/README.md)


jhthorsen:
Sounds about right, hehe

PR #482, opened 2020-05-27
When we change the theme, the color scheme select field is filled with correct values

thibaultduponchelle:
The color scheme list is correctly filled and the selected value is reset to 'auto'.

Main credit goes to @jhthorsen 

#477

### DIFF
--- a/assets/page/SettingsAccount.svelte
+++ b/assets/page/SettingsAccount.svelte
@@ -9,9 +9,25 @@ import {getContext} from 'svelte';
 import {l} from '../js/i18n';
 import {route} from '../store/Route';
 
-const colorSchemeOptions = [['auto', 'Auto'], ['dark', 'Dark'], ['light', 'Light']];
 const user = getContext('user');
 const updateUserOp = user.api.operation('updateUser');
+
+// "colorSchemeOptions" will get calculated each time "theme" changes
+$: colorSchemeOptions = calculateColorSchemeOptions(theme);
+$: colorScheme = resetColorScheme(theme);
+
+function calculateColorSchemeOptions(id) {
+  const options = [['auto', 'Auto']];
+  const theme = user.themes[id] || {variants: {}};
+  if (theme.variants.dark) options.push(['dark', 'Dark']);
+  if (theme.variants.light) options.push(['light', 'Light']);
+  return options;
+}
+
+function resetColorScheme(id) {
+  colorScheme = 'auto';
+  return colorScheme;
+}
 
 const themes = Object.keys(user.themes).map(id => {
   let name = user.themes[id].name;
@@ -20,8 +36,9 @@ const themes = Object.keys(user.themes).map(id => {
   return [id, name];
 });
 
+
 let formEl;
-let colorScheme = user.colorScheme;
+colorScheme = user.colorScheme;
 let expandUrlToMedia = user.embedMaker.expandUrlToMedia;
 let notificationsDisabled = user.omnibus.notifyPermission == 'denied';
 let theme = user.theme;
--- a/assets/page/SettingsAccount.svelte
+++ b/assets/page/SettingsAccount.svelte
@@ -12,23 +12,6 @@ import {route} from '../store/Route';
 const user = getContext('user');
 const updateUserOp = user.api.operation('updateUser');
 
-// "colorSchemeOptions" will get calculated each time "theme" changes
-$: colorSchemeOptions = calculateColorSchemeOptions(theme);
-$: colorScheme = resetColorScheme(theme);
-
-function calculateColorSchemeOptions(id) {
-  const options = [['auto', 'Auto']];
-  const theme = user.themes[id] || {variants: {}};
-  if (theme.variants.dark) options.push(['dark', 'Dark']);
-  if (theme.variants.light) options.push(['light', 'Light']);
-  return options;
-}
-
-function resetColorScheme(id) {
-  colorScheme = 'auto';
-  return colorScheme;
-}
-
 const themes = Object.keys(user.themes).map(id => {
   let name = user.themes[id].name;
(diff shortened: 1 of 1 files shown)


assets/page/SettingsAccount.svelte
```diff
@@ -9,19 +9,36 @@ import {getContext} from 'svelte';
 import {l} from '../js/i18n';
 import {route} from '../store/Route';
 
-const colorSchemeOptions = [['auto', 'Auto'], ['dark', 'Dark'], ['light', 'Light']];
 const user = getContext('user');
 const updateUserOp = user.api.operation('updateUser');
 
+// "colorSchemeOptions" will get calculated each time "theme" changes
```
jhthorsen, line 15: We don't really need this comment, it was just me trying to explain what this line does.

assets/page/SettingsAccount.svelte
```diff
@@ -9,19 +9,36 @@ import {getContext} from 'svelte';
...
+
+function calculateColorSchemeOptions(id) {
+  const options = [['auto', 'Auto']];
+  const theme = user.themes[id] || {variants: {}};
+  if (theme.variants.dark) options.push(['dark', 'Dark']);
+  if (theme.variants.light) options.push(['light', 'Light']);
+  return options;
+}
+
+function resetColorScheme(id) {
```
jhthorsen, line 27: I think this whole function is redundant: It assigns to `colorScheme`, but also returns `colorScheme` which earlier gets assigned to `colorScheme`. Maybe you would rather merge `calculateColorSchemeOptions()` and `resetColorScheme()`, like this:

```javascript
$: calculateColorSchemeOptions(theme);

function calculateColorSchemeOptions(id) {
  const options = [['auto', 'Auto']];
  const theme = user.themes[id] || {variants: {}};
  if (theme.variants.dark) options.push(['dark', 'Dark']);
  if (theme.variants.light) options.push(['light', 'Light']);
  colorSchemeOptions = options;
  colorScheme = 'auto';
}
```

assets/page/SettingsAccount.svelte
```diff
@@ -9,19 +9,36 @@ import {getContext} from 'svelte';
...
   let name = user.themes[id].name;
   const colorSchemes = Object.keys(user.themes[id].variants || {}).filter(cs => cs != 'default');
   if (colorSchemes.length) name += ' (' + colorSchemes.sort().join('/') + ')';
   return [id, name];
 });
 
+
 let formEl;
-let colorScheme = user.colorScheme;
+colorScheme = user.colorScheme;
```
jhthorsen, line 41: With the suggested `calculateColorSchemeOptions()` change above, then this must be changed back to `let ...`, which is the right thing to do, since it is bound to the SelectField component.

thibaultduponchelle: Done

assets/page/SettingsAccount.svelte
thibaultduponchelle: Done

assets/page/SettingsAccount.svelte
thibaultduponchelle: Removed

thibaultduponchelle:
Updated after your feedback and tested again

assets/page/SettingsAccount.svelte
```diff
@@ -9,18 +9,30 @@ import {getContext} from 'svelte';
...
 const themes = Object.keys(user.themes).map(id => {
   let name = user.themes[id].name;
   const colorSchemes = Object.keys(user.themes[id].variants || {}).filter(cs => cs != 'default');
   if (colorSchemes.length) name += ' (' + colorSchemes.sort().join('/') + ')';
   return [id, name];
 });
 
+
 let formEl;
+let colorSchemeOptions;
```
jhthorsen, line 35: Does this work without a default value? It doesn't have to be set to `[]` initially..?

assets/page/SettingsAccount.svelte
```diff
@@ -9,18 +9,30 @@ import {getContext} from 'svelte';
 import {l} from '../js/i18n';
 import {route} from '../store/Route';
 
-const colorSchemeOptions = [['auto', 'Auto'], ['dark', 'Dark'], ['light', 'Light']];
 const user = getContext('user');
 const updateUserOp = user.api.operation('updateUser');
 
+$: calculateColorSchemeOptions(theme);
+
+function calculateColorSchemeOptions(id) {
```
jhthorsen, line 17: Sorry for the nitpicking, but could you move this down to the rest of the functions and in alphabetic order?

assets/page/SettingsAccount.svelte
```diff
@@ -9,18 +9,30 @@ import {getContext} from 'svelte';
...
+}
+
 const themes = Object.keys(user.themes).map(id => {
   let name = user.themes[id].name;
   const colorSchemes = Object.keys(user.themes[id].variants || {}).filter(cs => cs != 'default');
   if (colorSchemes.length) name += ' (' + colorSchemes.sort().join('/') + ')';
   return [id, name];
 });
 
+
```
jhthorsen, line 33: Extra newline.

assets/page/SettingsAccount.svelte
thibaultduponchelle: It worked without value, but I just added it as you suggested

assets/page/SettingsAccount.svelte
thibaultduponchelle: Removed

assets/page/SettingsAccount.svelte
thibaultduponchelle: Done

jhthorsen:
Regarding the credits: Most of the work is checking that a new feature works, so many many thanks for the code change and the testing @thibaultduponchelle!

PR #484, opened 2020-05-29
Update tests after the recent addition of "Hacker" new theme.

thibaultduponchelle:


jhthorsen:
You beat me to it! 👍 

I'm going to take this, even if it's not quite right:

1. We use "perltidy" to make the code look the same everywhere, but I'm not going to force you to install it. If you want it for later though, then I would suggest installing https://metacpan.org/pod/App::githook::perltidy. Please ask me on IRC, if you want to install this tool.
2. The keys are out of order. "hacker" should be further up.

PR #492, opened 2020-06-01
Reduce docker image size.

marcusramberg:
Run `apk add`, `convos install` and cleanup in one RUN statement
to avoid extra layer creation. Reduces the image size from 269
to 59 MB:

```
marcus/convos       latest     33235db3041d        6 seconds ago       59MB
nordaaker/convos    latest     6edb9449fdce        9 hours ago         296MB
```


jhthorsen:
Looks awesome 👍

PR #494, opened 2020-06-05
Add perl-module-install to Dockerfile so Text::Markdown installs

stigtsp:


jhthorsen:
Thanks for the PR @stigtsp, but I think this might have been fixed in 1e4585bbdc588a25cbf8671f4ee626a772fbc42a?

stigtsp:
@jhthorsen Yep, seems like it. But does that make Module::Install a runtime dependency?

stigtsp:
It wouldn't be a problem for most cases. But one case would be for nixpkgs where tests will succeed without Module::Install, but fail runtime if the module isn't declared as a dependency for the closure, I've hacked in a test for it in the build https://github.com/NixOS/nixpkgs/pull/88940/files#diff-01e16fbbd11c9a1dddb0abc99d8c59efR40-R41

PR #506, opened 2020-07-23
Fix exif rotation

marcusramberg:
According to this gist
https://gist.github.com/runeb/c11f864cd7ead969a5f0 exif rotation is
required to handle endianness to work on all files. Tested this change
locally and it rotated the file from #convos earlier today correctly.

### DIFF
--- a/assets/js/EmbedMaker.js
+++ b/assets/js/EmbedMaker.js
@@ -149,16 +149,21 @@ export default class EmbedMaker extends Reactive {
 
     let maxBytes = dv.byteLength;
     let pos = 2;
+    var littleEndian = false;
     while (pos < maxBytes - 2 && !img.orientation) {
       const uint16 = dv.getUint16(pos);
       pos += 2;
       switch (uint16) {
         case 0xffe1: // Start of EXIF
-          maxBytes = dv.getUint16(pos) + pos;
+          var endianNess = scanner.getUint16(idx + 8);
+          if (endianNess === 0x4949) {
+            littleEndian = true;
+          }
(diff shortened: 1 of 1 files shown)


assets/js/EmbedMaker.js
```diff
@@ -149,16 +149,21 @@ export default class EmbedMaker extends Reactive {
...
       const uint16 = dv.getUint16(pos);
       pos += 2;
       switch (uint16) {
         case 0xffe1: // Start of EXIF
-          maxBytes = dv.getUint16(pos) + pos;
+          var endianNess = scanner.getUint16(idx + 8);
+          if (endianNess === 0x4949) {
+            littleEndian = true;
+          }
+        maxBytes = dv.getUint16(pos,littleEndian)-pos;
```
jhthorsen, line 162: Indentation is wrong and I would appreciate a more space:

```
maxBytes = dv.getUint16(pos, littleEndian) - pos;
```

assets/js/EmbedMaker.js
```diff
@@ -149,16 +149,21 @@ export default class EmbedMaker extends Reactive {
 
     let maxBytes = dv.byteLength;
     let pos = 2;
+    var littleEndian = false;
```
jhthorsen, line 152: "var" is not what the cool kids use anymore, please replace it with `let` and I would also sort the variables by name.

assets/js/EmbedMaker.js
jhthorsen: This is still not correct. I also wonder why it's not picked up by eslint...

jhthorsen:
I'm going to make a new release now, so I'll merge it and fix the white-space.

PR #509, opened 2020-07-30
Remove ! prefix from public commands

marcusramberg:
----

# Remove ! prefix

I didn't find any documentation or tests that were affected by 
this change. I think the user experience is better without the
! prefix, but as an alternative, we could make the prefix configurable
and default to an empty prefix.


jhthorsen:
I think I simply forgot to ask about “!” before I committed the bot feature. Wasn’t sure if it was a good idea in the first place.

Thanks for removing it 👍

PR #514, opened 2020-09-17
Force LF line ending for executable Perl scripts (fix Docker on Windows)

darktohka:
When building the Dockerfile with default Git settings on Docker for Windows, the build fails right after installing dependencies:

```
...
(19/21) Installing perl-utils (5.30.3-r0)
(20/21) Installing perl-dev (5.30.3-r0)
(21/21) Installing builddeps (20200917.205502)
Executing busybox-1.31.1-r9.trigger
OK: 225 MiB in 43 packages
': No such file or directory
The command '/bin/sh -c apk add --no-cache perl perl-io-socket-ssl wget &&     apk add --no-cache --virtual builddeps build-base perl-dev &&     /app/script/convos install --all &&     apk del builddeps && rm -rf /root/.cpanm /var/cache/apk/*' returned a non-zero code: 127
```

This is because when /bin/sh runs `/app/script/convos`, it opens the script and looks for a shebang line to figure out which executable to run the script with.

In other words, it's looking for `#!/usr/bin/env perl`.

However, since `/app/script/convos` is using CRLF line endings, `/bin/sh` does not know how to find the shebang, and therefore it doesn't know how to execute `/app/script/convos`.

This commit adds a `.gitattributes` file which forces Git to checkout all files inside the "script" directory with LF line endings. This allows the build script to continue successfully.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+script/* eol=lf


jhthorsen:
I don't understand this. In macos, I ran this command:

```
$ head -n2 script/convos | hexdump -C
00000000  23 21 2f 75 73 72 2f 62  69 6e 2f 65 6e 76 20 70  |#!/usr/bin/env p|
00000010  65 72 6c 0a 42 45 47 49  4e 20 7b 0a              |erl.BEGIN {.|
0000001c
```

...and it only show "0a" (LF) as the last character and not "0d0a" (CRLF).

Can you tell me where the CR comes from? Does git-for-windows change the newlines when checking out the source code?

darktohka:
Yes, on Windows the default Git behavior is to checkout files as CRLF.

jhthorsen:
Out of curiosity: How does that work when the files are checked in again? Will that create a huge diff or does git convert it back?

darktohka:
When commiting files, Git commits them with LF instead of CRLF. Since these files are already committed into the repo using LF, there won't be a huge diff if the file is edited after this commit is pushed.

Git takes care of the conversion automatically.

PR #515, opened 2020-10-03
Update ConnectionForm.svelte

wylel:
Add text for clarification on multiple commands in on-connect commands.


jhthorsen (approved):
I was looking for a “help” slot, but couldn’t find it, so I’ll take this PR.

I might change (unless someone beats me to it) the TextArea component so it can take a help slot, like this: https://github.com/Nordaaker/convos/blob/00a95979a13762b502b07ff6d8d7257247ceda0c/assets/components/form/TextField.svelte#L25

jhthorsen:
Thank you. Maybe a small change, but an important one 👍

PR #522, opened 2020-11-20
Ensure markdown link text start with a letter

marcusramberg:
Make link generation more strict to avoid accidentally triggering on regular expressions and such.

Was wondering if we should allow white space before and/or to start with a number, but I couldn't come up with any good use cases.


jhthorsen (approved):
I think this is a great start 👍 

(Still would like to disable substitutions inside backticks though)

PR #524, opened 2020-11-21
Chat input is coupled to conversation and restored when we go back from settings

thibaultduponchelle:


assets/store/Conversation.js
```diff
@@ -53,6 +53,7 @@ export default class Conversation extends Reactive {
 
     if (params.conversation_id) {
       this.prop('persist', 'wantNotifications', this.is_private, {key: params.conversation_id +  ':wantNotifications'});
+      this.prop('rw', 'input', params.input || '');
```
jhthorsen, line 56: Maybe we can call it "keyboardInput" or "userInput"? Naming it "input" makes it bit hard to understand what it's really used for.

Also no need to do read default from `params`. Just default to `''`.

assets/components/ChatInput.svelte
```diff
@@ -92,6 +95,9 @@ export function sendMessage(e) {
   const msg = {method: 'send'};
   msg.message = e.preventDefault ? inputEl.value : e.message;
 
+  conversation.update({input:''});
```
jhthorsen, line 98: Please add a space: `{input: ''}`.

assets/components/ChatInput.svelte
```diff
@@ -25,6 +26,8 @@ $: inputParts = calculateInputParts(pos);
 $: nick = connection && connection.nick;
 $: placeholder = conversation.is('search') ? 'What are you looking for?' : connection && connection.is('unreachable') ? l('Connecting...') : l('What is on your mind %1?', nick);
 $: sendIcon = conversation.is('search') ? 'search' : 'paper-plane';
+$: pos && (conversation.update({input: inputEl.value}));
```
jhthorsen, line 29: No need for parenthesis around conversation.update().

jhthorsen (changes requested):
I'm very positive to this change. Just needs some small changes.

Great work 👍

assets/store/Conversation.js
thibaultduponchelle: Fixed (userInput)

assets/components/ChatInput.svelte
thibaultduponchelle: Fixed

assets/components/ChatInput.svelte
thibaultduponchelle: Fixed

assets/store/Conversation.js
```diff
@@ -53,6 +53,7 @@ export default class Conversation extends Reactive {
 
     if (params.conversation_id) {
       this.prop('persist', 'wantNotifications', this.is_private, {key: params.conversation_id +  ':wantNotifications'});
+      this.prop('rw', 'userInput', '');
```
jhthorsen, line 56: If you want to make it persistent (meaning the input value survives a browser refresh), you can do something like this:

```
this.prop('rw', 'userInput', '', {key: ['userInput', params.connection_id, params.conversation_id].join(':')});
```

"key" is the value key used to store the value in [localStorage](https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage). It will default to the prop name, but you have to override it to make it unique per Conversation object.

jhthorsen:
Merging this after talking to @thibaultduponchelle on IRC.

Issue #380, opened 2019-03-03
Add nickserv certfp authentication


Svetlana-T:
https://freenode.net/kb/answer/certfp

https://www.oftc.net/NickServ/CertFP/

Would be nice to have, as some channels require authentication prior to joining and for the lack of a working automatic identification option, currently I have to join them by hand.

Wikiwide:
Would be nice. Especially taking into account that I regularly see disconnections (for whatever reasons), and client, annoyingly, is unable to automatically rejoin the channel.

jhthorsen:
Does that mean we should have two new fields in the connection settings, where you can upload a private and public key? If, so I think the building blocks are in place:

* 69213a66 has logic for uploading files that should only be accessible from the server.
* 26bbf628d has logic for reading tls_cert and tls_key, which could come from the files uploaded.

jhthorsen:
This should be implemented in 5.07.

Issue #356, opened 2017-07-16
SASL support connecting to a remote irc server


NinjaBanjo:
Any chance of getting SASL support? freenode doesn't allow you to connect from blacklisted ip ranges without SASL auth (behind a vpn/proxy)

marcusramberg:
Not sure how easy it would be for us to do. Just leaving this here as initial documentation of the IRC SASL protocol. http://ircv3.net/specs/extensions/sasl-3.1.html

jhthorsen:
I'm going to close this issue, since I don't think there's enough interest to get it implemented.

I don't mind a PR though :)

jhthorsen:
Going to re-open this issue, since it seems like Freenode blacklist AWS IP blocks. (Information from #convos on freenode)

PR #525, opened 2020-11-21
Convos::Core::Connection will generate custom certs for tls connections

jhthorsen:
This PR will generate custom certificates for each connection. I'm pretty sure this is required for #356 and #380.

### DIFF
--- a/lib/Convos/Core/Connection.pm
+++ b/lib/Convos/Core/Connection.pm
@@ -2,13 +2,15 @@ package Convos::Core::Connection;
 use Mojo::Base 'Mojo::EventEmitter';
 
 use Convos::Core::Conversation;
-use Convos::Util qw(DEBUG has_many);
+use Convos::Util qw(DEBUG generate_cert_p has_many);
 use Mojo::JSON qw(false true);
 use Mojo::Loader 'load_class';
 use Mojo::Promise;
 use Mojo::URL;
 use Mojo::Util qw(term_escape url_escape);
 use Unicode::UTF8;
+
+use constant GENERATE_CERT => $ENV{CONVOS_GENERATE_CERT} // '1';
 
 $IO::Socket::SSL::DEBUG = $ENV{CONVOS_TLS_DEBUG} if $ENV{CONVOS_TLS_DEBUG};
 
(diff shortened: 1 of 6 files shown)


stigtsp (commented):
Some thoughts:

Do we need a separate CA for this purpose?

And, `_openssl_detect` reads openssl.cnf/%ENV and set misc certificate parameters, this could be a bit overkill in the case of Freenode. They say: "You will be prompted for various pieces of information about the certificate. The   contents do not matter for our purposes, but openssl needs at least one of them to be non-empty."

Creating a self signed certificate using OpenSSL can be accomplished with an openssl oneliner:

  `openssl req -new -sha256 -newkey rsa:4096 -days 3650 -nodes -x509 -subj "/CN=$nick/O=Convos" -keyout self_signed_client.key -out self_signed_client.crt`

If `rsa` is hard coded, we should default the key size to 3072 or 4096 bits. https://www.keylength.com/

Just to be sure, is this the correct flow we're going for?
1. Generate a selfsigned cert for the connection
2. Connect to the irc server with the client certificate
3. Log in to NickServ
4. Do /msg NickServ CERT ADD to pin the self signed cert

jhthorsen:
> Do we need a separate CA for this purpose?
I don't know to be honest. I hope not, since I think that would make the process a lot more difficult.

> And, `_openssl_detect` reads openssl.cnf/%ENV and set misc certificate parameters, this could be a bit overkill in the case of Freenode. They say: "You will be prompted for various pieces of information about the certificate. The contents do not matter for our purposes, but openssl needs at least one of them to be non-empty."
Ah, I see. That makes the code a lot simpler 👍 

> If `rsa` is hard coded, we should default the key size to 3072 or 4096 bits. https://www.keylength.com/
Let's go for 4096.

> Just to be sure, is this the correct flow we're going for?
I think that is the flow for #380. I haven't been able to try out #356. There's some *buggy* code in https://github.com/Nordaaker/convos/tree/jhthorsen/sasl-certp, but I was not able to connect with SASL.

jhthorsen:
I think I've done the changes @stigtsp suggested now in a8533ef12a5e3500be388060027afc6455718c36.

Thanks for the input!

stigtsp:
Cool! :+1: Sorry for not being able to contribute with code for this PR :-/

PR #527, opened 2020-11-21
New setting to ignore statuses (john joined/john parted)

thibaultduponchelle:


assets/page/SettingsAccount.svelte
```diff
@@ -30,6 +31,7 @@ updateUserOp.on('start', req => {
   if (!req.body.password) delete req.body.password;
   req.body.highlight_keywords = req.body.highlight_keywords.split(/[.,\s]+/).map(str => str.trim());
   user.update({highlightKeywords: req.body.highlight_keywords});
+  user.update({ignoreStatuses: user.ignoreStatuses});
```
jhthorsen, line 34: I think this is a bug. Should probably just be `user.update({ignoreStatuses})`. You can also do this: `user.update({highlightKeywords: req.body.highlight_keywords, ignoreStatuses});` because of the magical `{ignoreStatuses} == {ignoreStatuses: ignoreStatuses}` syntax in es2015.

jhthorsen: Oh. Strike the previous comment. Just get rid of this, since you got it covered inside `updateUserFromForm()`.

assets/page/SettingsAccount.svelte
```diff
@@ -76,6 +79,10 @@ function updateUserFromForm(e) {
     <Checkbox name="notifications" bind:checked="{wantNotifications}">
       <span slot="label">{l('Enable notifications')}</span>
     </Checkbox>
+    
+    <Checkbox name="statuses" bind:checked="{ignoreStatuses}">
+      <span slot="label">{l('Ignore statuses')}</span>
```
jhthorsen, line 84: Can we make that easier to understand? Like "Ignore join/part messages" or something like that?

assets/store/Connection.js
```diff
@@ -122,7 +122,9 @@ export default class Connection extends Conversation {
   wsEventJoin(params) {
     const conversation = this.ensureConversation(params);
     const nick = params.nick || this.nick;
-    if (nick != this.nick) conversation.addMessage({message: '%1 joined.', vars: [nick]});
+    if ((nick != this.nick) && (!params.silent)) {
+        conversation.addMessage({message: '%1 joined.', vars: [nick]});
```
jhthorsen, line 126: You don't need all the parenthesis. Also two space indentation and not four.

assets/store/User.js
```diff
@@ -25,6 +25,7 @@ export default class User extends Reactive {
     this.prop('rw', 'highlightKeywords', []);
     this.prop('rw', 'rtc', {});
     this.prop('rw', 'status', 'pending');
+    this.prop('rw', 'ignoreStatuses', 1);
```
jhthorsen, line 28: I think the default should be changed from `1`to `false`. Also, changing this from "rw" to "persist" (no need for `key` here) will make the browser remember the setting.

assets/store/User.js
```diff
@@ -104,6 +105,7 @@ export default class User extends Reactive {
       default_connection: data.default_connection || '',
       forced_connection: data.forced_connection || false,
       highlightKeywords: data.highlight_keywords || [],
+      ignoreStatuses: data.ignoreStatuses || [],
```
jhthorsen, line 108: We are not able to handle `ignoreStatuses` in the backend yet, so this won't work. You either have to update the Perl code and API spec or remove this.

assets/store/User.js
```diff
@@ -155,6 +158,10 @@ export default class User extends Reactive {
     if (conversation && conversation[msg.dispatchTo]) conversation[msg.dispatchTo](msg);
   }
 
+  _shouldBeSilentMessage(msg) {
+    return ((msg.event == 'state') && this.ignoreStatuses);
```
jhthorsen, line 162: Oh. This was much simpler than what I had in mind. I guess you can just do this then:

```
msg.silent = this.ignoreStatuses;
// or
msg.ignoreStatuses = this.ignoreStatuses;
```

jhthorsen (changes requested):

assets/page/SettingsAccount.svelte
thibaultduponchelle: Fixed

PR #528, opened 2020-11-21
Show more info from WHOIS response by default

amenonsen:
We used to say "nick (host) is …"

Now we say "nick (user@host, ircname) is …"

(Also remove outdated comment about displaying whois response in current
conversation, which seems to work correctly now.)


jhthorsen (approved):
I think many users will be happy about this 👍

PR #529, opened 2020-11-21
Add "realname" as a new connection parameter

amenonsen:
You can use this to set the realname which is sent to the IRC server in
the initial USER command, and will be displayed in your WHOIS output.

Convos will append "via https://convos.chat" to the value you specify,
so as to have a simple and mostly-reliable way to detect other clients
using Convos.

### DIFF
--- a/assets/components/ConnectionForm.svelte
+++ b/assets/components/ConnectionForm.svelte
@@ -54,6 +54,7 @@ function defaultsToForm() {
   if (connURL.password) formEl.password.value = connURL.password;
   if (connURL.username) formEl.username.value = connURL.username;
   if (connParams.get('nick')) formEl.nick.value = connParams.get('nick');
+  if (connParams.get('realname')) formEl.realname.value = connParams.get('realname');
   if (connURL.pathname && formEl.conversation) formEl.conversation.value = decodeURIComponent(connURL.pathname.split('/').filter(p => p.length)[0] || '');
   if (Number(connParams.get('tls') || 0)) useTls = true;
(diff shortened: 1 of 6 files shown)


assets/components/ConnectionForm.svelte
```diff
@@ -110,10 +112,14 @@ async function submitForm(e) {
     </p>
   </TextField>
 
-  <TextField name="nick" placeholder="{l('Ex: your-name')}">
+  <TextField name="nick" placeholder="{l('Ex: superman')}">
     <span slot="label">{l('Nickname')}</span>
   </TextField>
 
+  <TextField name="realname" placeholder="{l('Ex: Clark Kent')}">
+    <span slot="label">{l('Real name')}</span>
```
jhthorsen, line 120: Is it less or more confusing with "Your name"? Also, do we need some help text to explain that this will be visible to other users you chat with?

jhthorsen (changes requested):
I would appreciate a test new test in https://github.com/Nordaaker/convos/commit/7408a0ee7239cca499d6ad1cc3da802469f8c97d as well.

assets/components/ConnectionForm.svelte
amenonsen: I think "superman/Clark Kent" is fine, but "nick/Your name" would be fine too. No strong preference.

How about "Real name (visible in WHOIS)"?

amenonsen: Changed as requested.

jhthorsen (approved):

PR #530, opened 2020-11-21
Clicking on a nick will create a new conversation

jhthorsen:
Before this PR, clicking on a nick would put that nick name into the chat input. After this PR it will...

1. _Clicking on your own nick will take you to "Account"_ -- Changed to same behavior as 2 for consistency
2. Clicking on some other nick will start a private conversation


amenonsen:
Tested against master, works as expected.

PR #533, opened 2020-11-23
Add an option to gather notifications per participant.

thibaultduponchelle:


thibaultduponchelle:
Since we discussed multiple topics on IRC, I'm not sure what is the status of this PR. I'm not sure but I feel like I have to close it without merge, do you agree ?

jhthorsen:
Unfortunately, I don't think this can be solved in the frontend.

thibaultduponchelle:
Ok :)

PR #535, opened 2020-11-25
Add a "compact display" mode as Viewport option

amenonsen:
Selecting this checkbox on the "account settings" page applies a new
.is-compact class to the body, which changes a few of the CSS rules
applied to messages in order to use less vertical space.

This could use more CSS tweaks, but I'm posting it for review since the functionality works.

### DIFF
--- a/assets/page/SettingsAccount.svelte
+++ b/assets/page/SettingsAccount.svelte
@@ -16,6 +16,7 @@ const user = getContext('user');
 const updateUserOp = api('updateUser');
 
 let formEl;
+let compactDisplay = viewport.compactDisplay;
 let expandUrlToMedia = viewport.expandUrlToMedia;
 let highlightKeywords = user.highlightKeywords.join(', ');
 let selectedColorScheme = viewport.colorScheme;
@@ -54,7 +55,7 @@ function updateUserFromForm(e) {
   }
 
   notify.update({wantNotifications});
-  viewport.update({expandUrlToMedia});
+  viewport.update({expandUrlToMedia, compactDisplay});
   user.update({ignoreStatuses});
   viewport.activateTheme(selectedTheme, colorSchemeReadonly ? '' : selectedColorScheme);
   updateUserOp.perform(e.target);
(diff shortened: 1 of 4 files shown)


assets/page/SettingsAccount.svelte
```diff
@@ -83,6 +84,10 @@ function updateUserFromForm(e) {
       <span slot="label">{l('Ignore join/part messages')}</span>
     </Checkbox>
 
+    <Checkbox name="compact" bind:checked="{compactDisplay}">
+      <span slot="label">{l('Enable compact message display')}</span>
+    </Checkbox>
```
jhthorsen, line 89: I think this checkbox should be moved under "Color scheme".

assets/store/Viewport.js
```diff
@@ -109,6 +110,17 @@ export default class Viewport extends Reactive {
...
   }
+
+  update(params) {
+    var body = document.getElementsByTagName('body')[0];
+    if (this.compactDisplay) {
+      body.classList.add('is-compact');
+    }
+    else {
+      body.classList.remove('is-compact');
+    }
```
jhthorsen, line 121: This can be golfed into `body.classList[this.compactDisplay ? 'add' : 'remove']('is-compact');`

assets/store/Viewport.js
```diff
@@ -109,6 +110,17 @@ export default class Viewport extends Reactive {
     if (typeof value == 'boolean') value = value ? 'yes' : 'no';
     el.content = value;
   }
+
+  update(params) {
+    var body = document.getElementsByTagName('body')[0];
```
jhthorsen, line 115: Please change `var` to `const`.

jhthorsen (changes requested):
I haven't tested this one out, but it looks very close to get merged.

Would you have merged it yourself @amenonsen? As in, do you like how it looks? Can you post a screenshot so it can be easier for others to get an idea about the change?

assets/page/SettingsAccount.svelte
amenonsen: Done.

assets/store/Viewport.js
amenonsen: Done.

assets/store/Viewport.js
amenonsen: Done.

jhthorsen (approved):

amenonsen:
After using it for a while, I'm happy with it. Here are a couple of screenshots:

DeepinScreenshot_select-area_20201126121926
DeepinScreenshot_select-area_20201126122005

Can you guess which one is condensed mode? ;-)

jhthorsen:
Thanks! This looks good to me and I think it will make many users happy 👍

PR #538, opened 2020-11-26
Allow leading uppercase characters in usernames

amenonsen:
This was just an oversight in the earlier commit, there's no problem
with using an initial uppercase character.


jhthorsen (changes requested):
Could you add an UPPERCASE character to https://github.com/Nordaaker/convos/blob/master/t/irc-connect-realname.t so we don’t accidentally break this in the future?

amenonsen:
Thanks, I've added another clause to `t/irc-connect-realname.t` as requested.

jhthorsen:
Thank you 👍

PR #549, opened 2021-01-13
Translate convos to Norwegian.

marcusramberg:
Support the fjord language.


jhthorsen (approved):
I think there are some hyphen bugs, but I'll merge it anyways, since I'm about to make a new release and I think this is good enough for now.

Maybe we can get a new PR from @arne or @alilles later on?

PR #552, opened 2021-01-19
Spanish translation updated

ivan-avalos:
Not sure why a lot of previous commits were inserted in this pull request, but here you have it!


jhthorsen (changes requested):
You can also translate the parts inside `[...]`. Example: `[settings]` might become something like `[ajustes]`?

The format is `[visible text](/some/url)`.

ivan-avalos:
> The format is `[visible text](/some/url)`.

Oh, I didn't know it was Markdown, I thought it was some .po special syntax.

assets/i18n/es.po
```diff
@@ -88,7 +88,7 @@ msgstr "Correo del administrador"
 
 #: assets/page/Help.svelte:58
 msgid "Any message starting with a-z is prefixed with \"`/quote`\", when sending from a connection conversation."
-msgstr "Any message starting with a-z is prefixed with \"`/quote`\", when sending from a connection conversation."
+msgstr "Cualquier mensaje que comience con a-z se le antepone \"`/quote`\", al enviar desde una conexión de conversación."
```
jjatria, line 91: I'm not sure what the end result of this should be like, but here are some observations:

* If "connection conversation" is a type of "conversation", then "conexión de conversación" is backwards. That said, "conversación de conexión" doesn't sound too natural to me either (nor does "connection conversation" for that matter). Could we maybe drop the "conexión" entirely?

* At the very least this seems to need an "a" at the beginning: "_a_ cualquier mensaje X se le antepone Y"

I feel like I'd need to know the context this is used in to make a better suggestion, but maybe a less-literal translation gives us more choices. Maybe something like

> Al enviar desde una conversación [de conexión?], cualquier mensaje que comience con a-z será precedido por \"`/quote`\"

?

ivan-avalos: "Connection conversation" is not a regular conversation, but rather the connection data and welcome message noise, where you can enter commands not related to any channel in specific.

jjatria: Ah, gotcha. In that case I guess "conversación de conexión" seems better to me at least.

jjatria (approved):

jhthorsen (approved):
Thanks for the changes!

PR #553, opened 2021-01-19
Translate new strings to Nårsk

marcusramberg:


jhthorsen (approved):
I'll take it, but I'm still hoping @alilles or @arne would review it one day they have time 😺

PR #564, opened 2021-02-06
Add support for deploying convos as a helm chart

marcusramberg:


jhthorsen:
What's the status on this PR @marcusramberg? I'm not sure I like adding it to a root folder (unless there's a reason for that), but other than that I'm indifferent, since I don't know Helm at all.

PR #580, opened 2021-03-05
Update commands.js

PeGaSuS-Coder:
- Added support for some other services such as MemoServ, BotServ, HostServ and OperServ
- Changed the rewrite from "/msg target" to "/quote target" which makes Convos aliases work out of the box on DALnet which forces the users to address themselves to services with "/msg NickServ@services.dal.net", e.g. Using "/quote" instead "/msg" solves that issue.


jhthorsen:
Thanks for the PR! Sorry for taking so long before merging.

Please sort the commands and squash the commits next time.

jhthorsen:
I remember why I didn't just take it now. It broke tests 🙈

PR #584, opened 2021-04-04
Add support for connection profiles

jhthorsen:
This PR adds Convos::Core::ConnectionSettings which holds default settings for a connection. The idea is to add this to the "Settings" section in Convos to manage common settings between connections.

It might even take over settings such as...
- [ ] Secure connection (TLS)
- [ ] Verify certificate (TLS)

Even further, I'm considering removing many settings in the connection editor, and replace it with a dropdown where a user can select Convos::Core::ConnectionSettings profiles instead.

This is currently a WIP branch.

1tzmjrLP6NPfbWY0


jhthorsen:
1tzmjrLP6NPfbWY0

jhthorsen:
This branch also has a new connection editor.

PR #593, opened 2021-05-23
Compatible with Mojolicious::Plugin::Webpack 1.00

jhthorsen:
This PR makes Convos compatible with the next version of Mojolicious::Plugin::Webpack.

https://github.com/jhthorsen/mojolicious-plugin-webpack/pull/13


assets/sass/components/_link-embedder.scss
```diff
@@ -15,7 +15,7 @@
 
 .embed iframe {
   max-width: 90vw;
-  max-height: (344 / 459 * 90vw);
+  max-height: (0.7494553377 * 90vw) // 344 / 459 = 0.7494553377
```
jhthorsen, line 18: Sass has deprecated "/" as division. Need to use math.div() instead, but multiply works well enough.

PR #594, opened 2021-05-25
Add GNOME dark theme

losuler:
## Description

I originally made this theme as I'm a GNOME user and wanted something that fitted in with it's theming. I used [Fractal](https://gitlab.gnome.org/GNOME/fractal) and `gtk3-widget-factory` with GNOME dark theming enabled for reference. I made this a few weeks ago and have been using it since and haven't noticed any issues. I apologise if the css isn't great.

I'm not sure if you want to keep the `color-scheme: dark` and `-dark.css` in the filename, but originally when I wrote this I'd planned to make a light theme too, since that is the default for GNOME, but I never got around to it.

## Screenshots

Screenshot from 2021-05-25 20-48-14
Screenshot from 2021-05-25 17-56-56
Screenshot from 2021-05-25 17-57-14

### DIFF
--- a/public/themes/gnome_color-scheme-dark.css
+++ b/public/themes/gnome_color-scheme-dark.css
@@ -29,7 +29,6 @@
   --input-color: #fff;
   --input-focus-outline: 2px solid #15539e;
   --input-disabled-bg: #2d2d2d;
-  --input-disabled-color: #D8DEE9;
   --chat-input-bg: #353535;
   --autocomplete-bg: #353535;
   --autocomplete-border: 1px solid #1b1b1b;
--- a/public/themes/gnome_color-scheme-dark.css
+++ b/public/themes/gnome_color-scheme-dark.css
@@ -30,7 +30,7 @@
   --input-focus-outline: 2px solid #15539e;
   --input-disabled-bg: #2d2d2d;
   --input-disabled-color: #D8DEE9;
-  --chat-input-bg: #2d2d2d;
+  --chat-input-bg: #353535;
   --autocomplete-bg: #353535;
   --autocomplete-border: 1px solid #1b1b1b;
   --autocomplete-focus-bg: #15539e;
@@ -43,10 +43,6 @@
   --sidebar-right-bg: var(--body-bg);
   --sidebar-right-text: var(--text-color);
   --syntax-hl-base-bg: #353535;
-}
-
-.chat-input {
-  background: #353535;
 }
 
 .sidebar-left__nav > h3 {
--- a/public/themes/gnome_color-scheme-dark.css
+++ b/public/themes/gnome_color-scheme-dark.css
@@ -1,0 +1,58 @@
+/*
+ * name: GNOME
+ * color-scheme: dark
+ */
+
+:root {
+  --body-bg: #2d2d2d;
+  --cms-bg: var(--body-bg);
+  --text-color: #fff;
+  --link-color: #3472bd;
+  --quote-color: #3472bd;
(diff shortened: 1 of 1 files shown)


jhthorsen:
That looks very nice 👍
Can you also provide a screenshot of the login screen? Would very much like to see how the form elements are rendered.

public/themes/gnome_color-scheme-dark.css
```diff
@@ -0,0 +1,58 @@
...
+  --sidebar-left-border-bottom: #1b1b1b;
+  --sidebar-left-search-placeholder-color: #fff;
+  --sidebar-left-search-focus-bg: #15539e;
+  --sidebar-right-bg: var(--body-bg);
+  --sidebar-right-text: var(--text-color);
+  --syntax-hl-base-bg: #353535;
+}
+
+.chat-input {
+  background: #353535;
```
jhthorsen, line 49: Can’t you use --chat-input-bg?

jhthorsen (changes requested):
That looks very nice 👍
Can you also provide a screenshot of the login screen? Would very much like to see how the form elements are rendered.

losuler:
Here's a screenshot of the login screen.

Screenshot from 2021-05-25 21-28-53

public/themes/gnome_color-scheme-dark.css
losuler: When I do that it results in the input bar becoming the same color as the background of the other chat messages (example below).

Screenshot from 2021-05-25 21-33-11

jhthorsen: I meant redefining chat-input-bg:
```
--chat-input-bg: #353535;
```

public/themes/gnome_color-scheme-dark.css
```diff
@@ -0,0 +1,58 @@
...
+  --code-bg: #292929;
+  --button-color: #ECEFF4;
+  --button-border: 1px solid #1b1b1b;
+  --button-focus-outline: 2px solid #1b1b1b;
+  --button-disabled-bg: #323232;
+  --button-disabled-color: #eceff4;
+  --button-disabled-border: #1b1b1b;
+  --button-danger-bg: #ae151c;
+  --button-danger-color: #fff;
+  --input-bg: #2d2d2d;
```
jhthorsen, line 27: From the posted screenshot for the login screen, I think this have to change. The input fields blend too much with the background and they just don't feel like an input field.

jhthorsen (changes requested):

public/themes/gnome_color-scheme-dark.css
jhthorsen: Here is one idea:

C96BC6B2-19E8-4541-8B65-3DF6A174860D

losuler: How does this look?

Screenshot from 2021-05-26 00-59-59

public/themes/gnome_color-scheme-dark.css
losuler: Ah, I'm not sure why I didn't do that. Fixed this in https://github.com/convos-chat/convos/pull/594/commits/188f1fba0aa723d410150d2e03f0d63d883058d4.

public/themes/gnome_color-scheme-dark.css
jhthorsen: I still don't think there's enough contrast, but I'll take it anyways. I think it's better to iterate later on 👍

PR #595, opened 2021-05-28
Fix very bright preformatted message background in "Hacker" theme

compeak:
This PR turns this

bright

into this

dark

Preformatted messaged will be less blinding but also not as distinguishable from other messages as before.

### DIFF
--- a/public/themes/hacker.css
+++ b/public/themes/hacker.css
@@ -54,7 +54,7 @@
   /* Keep as is */
   --error-color: #ee2222;
   --success-color: #338833;
-  --syntax-hl-base-bg: #fff;
+  --syntax-hl-base-bg: black;
 
   /* Listbox */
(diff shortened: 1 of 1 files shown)


PeGaSuS-Coder:
why not use something like `#8c8c8c` as background?  
Less brighten and distinguishable at the same time.  
Cheers!

compeak:
@PeGaSuS-Coder I tried some greys but black looks best IMHO.

A green border on the left side could be an option.

(I just wanted to fix the blinding effect, to be honest.)

jhthorsen:
I must agree with @PeGaSuS-Coder, but I think I'm going to take this change anyways just because it is indeed an improvement.

jhthorsen:
Too bad I forgot to merge this before I just made a release 😞 Will be part of 6.24 at least.

PR #612, opened 2021-07-09
Castaway/dismiss notifications in channel

castaway:
Store unread notification count per conversation object, assemble a count of those for the sidebar display. Dismiss / set to zero (for that channel) when a user visits a channel that has notifications.

I'm missing a sane way to test this in __tests__ as it requires both User + Notifications objs, any ideas?

### DIFF
--- a/assets/components/ChatSidebar.svelte
+++ b/assets/components/ChatSidebar.svelte
@@ -11,6 +11,7 @@ import {route} from '../store/Route';
 export let transition;
 
 const user = getContext('user');
+const notifications = user.notifications;
 
 let activeLinkIndex = 0;
 let filter = '';
@@ -19,7 +20,6 @@ let searchHasFocus = false;
 let visibleLinks = [];
 
 $: filterNav({filter}); // Passing "filter" in to make sure filterNav() is called on change
-$: notifications = $user.notifications;
 $: addConversationLink = '/settings/conversation?connection_id=' + ($user.activeConversation.connection_id || '');
 $: searchQuery = filter.replace(/^\//, '');
 $: if (navEl) clearFilter($route);
--- a/assets/components/ChatSidebar.svelte
+++ b/assets/components/ChatSidebar.svelte
@@ -11,7 +11,6 @@ import {route} from '../store/Route';
 export let transition;
 
 const user = getContext('user');
-const notifications = $user.notifications;
 
 let activeLinkIndex = 0;
 let filter = '';
@@ -20,6 +19,7 @@ let searchHasFocus = false;
 let visibleLinks = [];
 
 $: filterNav({filter}); // Passing "filter" in to make sure filterNav() is called on change
+$: notifications = $user.notifications;
 $: addConversationLink = '/settings/conversation?connection_id=' + ($user.activeConversation.connection_id || '');
 $: searchQuery = filter.replace(/^\//, '');
 $: if (navEl) clearFilter($route);
--- a/assets/page/Chat.svelte
+++ b/assets/page/Chat.svelte
@@ -39,6 +39,7 @@ let focusChatInput, fillIn, uploader;
 
 $: setConversationFromRoute(connection_id, conversation_id);
 $: setConversationFromUser($user);
+$: updateNotificationCount($user);
 $: messages.update({expandUrlToMedia: $user.expandUrlToMedia});
 $: notConnected = $conversation.frozen ? true : false;
 $: title = $conversation.title;
@@ -97,6 +98,11 @@ function showPopover(e) {
       if (!$popoverTarget) target.click();
     }
   }, 250);
+}
+
+function updateNotificationCount(user) {
+  user.activeConversation.update({notifications: 0});
+  user.updateNotificationCount();
 }
 </script>
 
--- a/assets/page/Chat.svelte
+++ b/assets/page/Chat.svelte
@@ -39,7 +39,6 @@ let focusChatInput, fillIn, uploader;
 
 $: setConversationFromRoute(connection_id, conversation_id);
 $: setConversationFromUser($user);
-$: updateNotificationCount($user);
 $: messages.update({expandUrlToMedia: $user.expandUrlToMedia});
 $: notConnected = $conversation.frozen ? true : false;
 $: title = $conversation.title;
@@ -82,6 +81,7 @@ function setConversationFromUser(user) {
   now = new Time();
   unsubscribe.conversation = conversation.subscribe(d => { conversation = d });
   unsubscribe.markAsRead = conversation.markAsRead.bind(conversation);
+  updateNotificationCount(user);
 
   onLoadHash = isISOTimeString(route.hash) && route.hash || '';
   if (onLoadHash) return conversation.load({around: onLoadHash});
--- a/lib/Convos/Controller/Conversation.pm
+++ b/lib/Convos/Controller/Conversation.pm
@@ -9,7 +9,7 @@ sub mark_as_read {
   my $conversation = $self->backend->conversation({});
 
   unless ($conversation) {
-    return $self->reply->errors([],                        401) unless $self->backend->user;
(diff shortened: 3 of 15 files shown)


assets/components/ChatSidebar.svelte
```diff
@@ -11,14 +11,14 @@ import {route} from '../store/Route';
 export let transition;
 
 const user = getContext('user');
-const notifications = $user.notifications;
```
jhthorsen, line 14: Why did you change this? I think `$user.notifications` was a typo in the original code though. It should just be `user.notifications`.

lib/Convos/Core/Backend.pm
```diff
@@ -75,7 +75,7 @@ sub _setup {
           my ($connection, $target, $msg) = @_;
 
           if ($msg->{highlight} and $target->id and !$target->is_private) {
-            $connection->user->{unread}++;
+            $target->inc_notifications_p->catch(sub { $self->_debug('inc_notifications %s FAIL %s', $target->id, shift) } );
```
jhthorsen, line 78: Since the class doesn't really do anything, you can change this to inc_notifications() instead:

```
sub inc_notifications { $_[0]->{notifications}++; $_[0] }
```

lib/Convos/Core/Conversation.pm
```diff
@@ -9,6 +9,7 @@ has name     => sub { Carp::confess('name required in constructor') };
 has password => '';
 has topic    => '';
 has unread   => 0;
+has notifications => 0;
```
jhthorsen, line 12: Is it easier to read `notifications` or `n_notifications` ?

jhthorsen (changes requested):
I think this is a great start! Thank you 👍 

I'm a bit unsure what `updateNotificationCount()` in Chat.svelte does, but I will look closer when I wake up in the morning.

Can you also get rid of the extra code in Convos::Core::User that is not in use anymore? And I guess the API needs some changes as well to actually clear the notifications.

assets/components/ChatSidebar.svelte
castaway: Ah I think that one was me trying to fix a different issue, I'll put it back

lib/Convos/Core/Conversation.pm
castaway: Either works - I otherwise pondered "notifications_count" .. to me its somewhat obvious from the default being 0 .. else we'd need to change all the "unread" attributes too.

jhthorsen: It's obvious here, but not so obvious other places in the code where it is used. Going to resolve this issue though, since "unread" needs changing as well.

castaway:
Updated with API notifications value resetting to 0 at appropriate points.

Not sure I got Mojo's promises quite right tho as I got this:

```
[2021-07-15 17:47:56.06484] [1821] [debug] [dsWZhdKQfyV9] Routing to controller "Convos::Controller::Notifications" and action "read"
[irc-libera] Save success. (/mnt/allthespace/usrsrc/perl/convos/castaway@desert-island.me.uk/irc-libera/connection.json)
[2021-07-15 17:47:56.06832] [1821] [error] Can't locate object method "clone" via package "Mojo::Collection" at /usr/src/perl/libs/convos/lib/perl5/Mojo/Promise.pm line 120.

[2021-07-15 17:47:56.06956] [1821] [debug] [dsWZhdKQfyV9] 500 Internal Server Error (0.007406s, 135.026/s)
```

lib/Convos/Controller/Notifications.pm
```diff
@@ -16,7 +16,12 @@ sub read {
   my $self = shift->openapi->valid_input or return;
   my $user = $self->backend->user        or return $self->reply->errors([], 401);
 
-  return $user->unread(0)->save_p->then(sub { $self->render(openapi => {}) });
+  Mojo::Promise->all($user->connections->map(sub {
+      my $conn = shift;
+      $conn->conversations->map( sub { shift->notifications(0); });
+      $conn->save_p();
+    })
+  )->then(sub { $self->render(openapi => {}) });
```
jhthorsen, line 24: I don't think you need this complexity, since the connection is saved on interval. (You should be rather unlucky if it doesn't get saved automatically) I suggest something like this:

```
$user->connections->map(sub { shift->conversations->map(notifications => 0) });
$self->render(openapi => {});
```

But the problem with the original code is that `then()` is called on the collection (return value from map) not the promise method all(). Pretty sure this would work:

```
Mojo::Promise->all(
  $user->connections->map(sub {
    my $conn = shift;
    $conn->conversations->map(sub { shift->notifications(0); });
    return $conn->save_p;
  })->each
)->then(sub { $self->render(openapi => {}) });
```

lib/Convos/Controller/Conversation.pm
```diff
@@ -14,6 +14,7 @@ sub mark_as_read {
   }
 
   $conversation->unread(0);
+  $conversation->notifications(0);
```
jhthorsen, line 17: This can also be chained:

```
$conversation->notifications(0)->unread(0);
```

lib/Convos/Core/Backend.pm
```diff
@@ -75,7 +75,7 @@ sub _setup {
           my ($connection, $target, $msg) = @_;
 
           if ($msg->{highlight} and $target->id and !$target->is_private) {
-            $connection->user->{unread}++;
+            $target->inc_notifications();
```
jhthorsen, line 78: 👍

lib/Convos/Core/Conversation.pm
```diff
@@ -35,7 +38,7 @@ sub _calculate_unread_p {
 
 sub TO_JSON {
   my ($self, $persist) = @_;
-  my %json = map { ($_, $self->$_) } qw(frozen name topic unread);
+  my %json = map { ($_, $self->$_) } qw(frozen name topic unread notifications);
```
jhthorsen, line 41: I forgot to mention something else: I would appreciate if "things" where sorted. This goes for methods and attributes as well.

Also, if you could run the code through `perltidy`, then it would be less work afterwards, and you would get mentioned with "git blame".

I'm going to take this change either way, but I would greatly appreciate it.

jhthorsen (changes requested):

lib/Convos/Controller/Notifications.pm
castaway: Thanks! Revised as suggested

lib/Convos/Core/Conversation.pm
castaway: Righto, I shall try and remember that (updated)

lib/Convos/Controller/Conversation.pm
castaway: done

jhthorsen (changes requested):
There’s many attributes/methods/keywords that isn’t sorted, and the code has not been run through perltidy. (I use https://metacpan.org/dist/App-githook-perltidy/view/bin/githook-perltidy)

assets/page/Chat.svelte
```diff
@@ -39,6 +39,7 @@ let focusChatInput, fillIn, uploader;
 
 $: setConversationFromRoute(connection_id, conversation_id);
 $: setConversationFromUser($user);
+$: updateNotificationCount($user);
```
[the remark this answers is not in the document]
castaway: This is the update that sets notifications to 0 for this channel, when the user visits it (so should be on load of this channel, not on every message)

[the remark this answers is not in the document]
castaway: I put it here as that's where setActiveConversation is, which I assumed happened once per channel / page visit.

jhthorsen:
I've rebased your commits and fixed a couple of things. The changes are in 48e904850c5abfb1093ef7098a05c731bfe83a5c and 96ef5a269c2a2c9aaeb92a72af7afcab2045343c

PR #615, opened 2021-08-25
Add SQLite backend to Convos

jhthorsen:
This is currently just a draft to show how a new backend might look. I don't know when I will have time to finish this, so help is greatly appreciated.

TODO:

* Add a database schema
* Make sure objects and messages are inflated and deflated properly. (Some columns might contain JSON)
* Duplicate the ::File tests, but with `$ENV{CONVOS_BACKEND} = 'Convos::Core::Backend::SQLite'`
* Write a migration script, so existing ::File users can convert to ::SQLite without loosing data
* Add a detection step to "convos install" which should default to SQLIte, if SQLite is available on the system

There's probably more...


brunoramoslu:
Is this still in the roadmap as a possible/interesting feature to implement?
Could work be picked up, or continued, from here?

jhthorsen:
It's not on the roadmap, since the file backend works pretty good for me. It would however open new possibilities (like better search capabilities) so it would be awesome if someone could finish it 👍

PR #617, opened 2021-09-12
Don't notify user on mentions if they are active in that channel

castaway:


jhthorsen (commented):
This is pretty much the same as I have in e324170dc, so I'm going to close this.

Thanks for noticing!

Issue #464, opened 2020-05-11
Docker ARM builds


kisst:
**Is your feature request related to a problem? Please describe.**
Would like to run convos in a Docker container on a RPi

**Describe the solution you'd like**
Setup docker hub to build multi arch setup

**Describe alternatives you've considered**
Installing it on the fly to the container, but that's not as nice

**Additional context**
Quite a few docker hub image already support it either by a separated tag, or natively in case of the official image, should be doable either way.

jhthorsen:
Is this something you can look into @marcusramberg ?

marcusramberg:
I've added it to my TODO, but I don't know much about what's required.

jhthorsen:
Does not seem like we know how to do this, so I moved the issue from the 4.xx milestone to Backlog.

I hope it doesn't go into the [fridge](/Nordaaker/convos/milestone/20?closed=1).

marcusramberg:
https://www.docker.com/blog/multi-arch-build-and-images-the-simple-way/ seems to have some good info on how to do this with GitHub Actions. Will investigate further.

jhthorsen:
Looks like this won't happen without external help, so I'm placing this issue in the fridge for now.

jhthorsen:
This is fixed in #618 🎉

PR #618, opened 2021-09-19
Add arm64 docker image build support

Banshee1221:
This commit adds support for building arm64 images (should also be easy to add other architectures) with GitHub Actions. (Issue reference: #464). This action will generate both amd64 and arm64 images.

It also swaps out the hardcoded "convos/convos" URL for the secret `DOCKER_HUB_USERNAME`/convos which simplifies forks.

I've deployed the resultant container on a Raspberry Pi4 with Ubuntu Server 21.04 and all appears to work just fine. Container here: https://hub.docker.com/r/eugenedb/convos

### DIFF
--- a/.github/workflows/dockerimage.yml
+++ b/.github/workflows/dockerimage.yml
@@ -18,16 +18,15 @@ jobs:
       run: |
         if echo $GITHUB_REF | grep -q '^refs/tags/v'; then
           VERSION=$(echo $GITHUB_REF | cut -d/ -f3);
-          TAG_NAME="convos/convos:$VERSION";
+          TAG_NAME="${{ secrets.DOCKER_HUB_USERNAME }}/convos:$VERSION";
         elif [ "$GITHUB_REF" = "refs/heads/docker" ]; then
-          TAG_NAME="convos/convos:alpha";
+          TAG_NAME="${{ secrets.DOCKER_HUB_USERNAME }}/convos:alpha";
         elif [ "$GITHUB_REF" = "refs/heads/master" ]; then
-          TAG_NAME="convos/convos:alpha";
+          TAG_NAME="${{ secrets.DOCKER_HUB_USERNAME }}/convos:alpha";
         else
-          TAG_NAME="convos/convos:stable";
+          TAG_NAME="${{ secrets.DOCKER_HUB_USERNAME }}/convos:stable";
         fi
         echo '::set-output name=tag::'$TAG_NAME
-
     - name: Check Out Repo
       uses: actions/checkout@v1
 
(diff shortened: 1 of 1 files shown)


jhthorsen:
Looks great! Thank you 👍

jhthorsen:
This dramatically slowed down the build. Last build was less than four minutes, while it’s still running now after 10 minutes. I wonder if we could speed things up by creating about GitHub actions file?

Banshee1221:
@jhthorsen Yeah, I should have probably mentioned. My apologies. This builds the ARM image using emulation for cross-compilation. It will be quite a bit slower. 

To clarify, are you suggesting splitting out the action file for multiple architectures? That may be doable, but is it really necessary for the builds to be that quick?

Banshee1221:
I see I've also made a mistake assuming that your Dockerhub username is the same as the repo/org name. 

Can I go ahead and include a new secret for the action file `DOCKER_HUB_URL`, that would point the tag to the correct URL? I'll make a new pull request for that now as a quick fix and then have a look at making things quicker for builds later.

jhthorsen:
> That may be doable, but is it really necessary for the builds to be that quick?

I guess "that quick" is relative, but going from four minutes to forty minutes is quite the difference.

https://github.com/convos-chat/convos/actions/runs/1251648371 == 38m 54s
https://github.com/convos-chat/convos/actions/runs/1228301541 == 3m 37s

If possible, I would like them to be in the same action file, but ran twice: One for AMD64 and one for ARM. Meaning I hope https://github.com/convos-chat/convos/actions/runs/1251648371/workflow#L53 can be a variable that the action file loops over.

It's still crazy slow, but at least I get my own image in reasonable time. Everyone on arm64 won't care, since they aren't watching the screen while the image is building.

Banshee1221:
I'll dig into it

jhthorsen:
I fixed this in 32ca942a7441c02d48634cd284e7c648af6f87ba now. I would very much appreciate if you can look into how to build the arm64 image quicker though. 40 minutes is very, very slow 😞

Banshee1221:
Thanks @jhthorsen I also thought about breaking it out in a matrix. I'm digging into it a little and will see if we can opt for cross-compiling somehow instead of emulation (what the docker github action does by default), but will only be able to look into it after the start of October as I'll be away till then..

Banshee1221:
@jhthorsen just noticed that doing the changes you've made to the action results in the amd64 build being overwritten on Dockerhub with the arm64 build. It appears you have to build them together with buildx in order for the manifest to be properly created (listing both image architectures). I'll investigate more this week coming...

jhthorsen:
Yeah. This is a big problem :( I changed it now so that the amd64 image will have the proper version number, while arm64 will just be “arm64” now. Let me know if you have a better way to do this, without me back to wait 40 minute for the amd image

jhthorsen:
The builds are now fast again. The fix was to add a base image where most of the dependencies are already built: https://github.com/convos-chat/docker-base/

Banshee1221:
Excellent, I apologise for not being able to get back to you sooner. I'm really happy it got worked out. Super excited about this :)

PR #619, opened 2021-09-20
Add Dockerhub prefix variable for tag in Github action

Banshee1221:
This will fix my assumption in #618 that the Dockerhub username is the same name used as the Docker package prefix.

This requires a new secret `DOCKER_HUB_PREFIX` to be added to the repository. In this case `DOCKER_HUB_PREFIX`=`convos`.

Apologies for not realising this sooner!

### DIFF
--- a/.github/workflows/dockerimage.yml
+++ b/.github/workflows/dockerimage.yml
@@ -18,16 +18,15 @@ jobs:
       run: |
         if echo $GITHUB_REF | grep -q '^refs/tags/v'; then
           VERSION=$(echo $GITHUB_REF | cut -d/ -f3);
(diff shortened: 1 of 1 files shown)


jhthorsen:
Looks sane 👍

jhthorsen (changes requested):
Please rebase your changes. No reason to have three commits for this.

Looks good otherwise!

Note to merger: I've added the secret to this repo, so it's safe to merge this later on.

Banshee1221:
It appears that my Git skills are not the best!

jhthorsen (approved):
Ok, I'll squash it on github then 👍

PR #627, opened 2021-11-09
Use Crypt::Passphrase for password hashing

stigtsp:
- Replace _bcrypt with Crypt::Passphrase
- Default to argon2id with validation support for bcrypt

### DIFF
--- a/lib/Convos/Core/User.pm
+++ b/lib/Convos/Core/User.pm
@@ -3,17 +3,11 @@ use Mojo::Base 'Mojo::EventEmitter';
 
 use Convos::Core::Connection;
 use Convos::Util qw(DEBUG has_many);
-use Crypt::Eksblowfish::Bcrypt ();
+use Crypt::Passphrase;
 use File::Path                 ();
 use Mojo::Date;
 use Mojo::Promise;
 use Mojo::Util qw(camelize trim);
-
-use constant BCRYPT_BASE_SETTINGS => do {
-  my $cost = sprintf '%02i', 8;
-  my $nul  = 'a';
-  join '', '$2', $nul, '$', $cost, '$';
-};
 
 sub core  { shift->{core}  or die 'core is required in constructor' }
 sub email { shift->{email} or die 'email is required in constructor' }
@@ -24,6 +18,13 @@ has remote_address => '127.0.0.1';
 has roles          => sub { +[] };
 has uid            => sub { die 'uid() cannot be built' };
 has unread         => sub {0};
+
+has _crypt => sub {
+  Crypt::Passphrase->new(
+    encoder => { module => 'Argon2', memory_cost => '64M' },
+    validators => [ 'Bcrypt' ]
+  );
+};
 
 has_many connections => 'Convos::Core::Connection' => sub {
   my ($self, $attrs) = @_;
@@ -141,7 +142,7 @@ sub save_p {
 sub set_password {
   my ($self, $plain) = @_;
   die 'Usage: Convos::Core::User->set_password($plain)' unless $plain;
-  $self->{password} = $self->_bcrypt($plain);
+  $self->{password} = $self->_crypt->hash_password($plain);
   $self;
 }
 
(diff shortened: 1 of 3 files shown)


t/user.t
```diff
@@ -41,8 +41,13 @@ subtest 'password' => sub {
   eval { $user->set_password('') };
   like $@, qr{Usage:.*plain}, 'set_password() require plain string';
   ok !$user->password, 'no password';
+
+  $user->{password} = '$2a$08$0KlK5QzJjzWFNW6JsuT52..GXe1sTRWZU1es8hfo0HcD29tTFzvsi';
+  $user->validate_password('s3cret', 'validate password backwards compatible with bcrypt');
```
jhthorsen, line 46: ❤️

lib/Convos/Core/User.pm
```diff
@@ -151,21 +155,11 @@ sub validate_password {
   my ($self, $plain) = @_;
 
   return 0 unless $self->password and $plain;
-  return 1 if $self->_bcrypt($plain, $self->password) eq $self->password;
+  return 1 if $self->_crypt->verify_password($plain, $self->password);
+  # XXX: Should this also upgrade hash if $self->_crypt->needs_rehash($plain) ?
```
jhthorsen, line 159: I think it would be a great addition if the password would get upgraded automatically.

lib/Convos/Core/User.pm
```diff
@@ -3,18 +3,12 @@ use Mojo::Base 'Mojo::EventEmitter';
...
 use File::Path                 ();
 use Mojo::Date;
 use Mojo::Promise;
 use Mojo::Util qw(camelize trim);
 
-use constant BCRYPT_BASE_SETTINGS => do {
-  my $cost = sprintf '%02i', 8;
-  my $nul  = 'a';
-  join '', '$2', $nul, '$', $cost, '$';
-};
```
jhthorsen, line 16: It's really nice to see this is gone 👍

lib/Convos/Core/User.pm
```diff
@@ -25,6 +19,16 @@ has roles          => sub { +[] };
...
 
+has _crypt => sub {
+  Crypt::Passphrase->new(
+    encoder => {
+      module => 'Argon2',
+      memory_cost => '64M'
+    },
+    validators => [ 'Bcrypt' ]
+  );
+};
```
jhthorsen, line 30: I don't think this is run through perltidy. Should look something like this:

```
has _crypt => sub {
  Crypt::Passphrase->new(
    encoder    => {module => 'Argon2', memory_cost => '64M'},
    validators => ['Bcrypt']
  );
};
```

jhthorsen (changes requested):
This is just great! Thank you.

lib/Convos/Core/User.pm
stigtsp: This is fixed.

(Not sure if calling save_p is neccecary?)

t/user.t
stigtsp: This test was broken, woopsie, so fixed it.

lib/Convos/Core/User.pm
```diff
@@ -151,19 +152,11 @@ sub validate_password {
...
-  unless ($settings) {
-    my $salt = join '', map { chr int rand 256 } 1 .. 16;
-    $settings = BCRYPT_BASE_SETTINGS . Crypt::Eksblowfish::Bcrypt::en_base64($salt);
+  my $valid = $self->_crypt->verify_password($plain, $self->password);
+  if ($valid && $self->_crypt->needs_rehash($plain)) {
+    $self->{password} = $self->_crypt->hash_password($plain);
   }
-
-  Crypt::Eksblowfish::Bcrypt::bcrypt($plain, $settings);
+  return $valid;
```
stigtsp, line 159: This now returns nothing rather than 0, which I think should be ok (and maybe even a little bit safer).

jhthorsen: Why is it safer?

stigtsp:
Was thinking if it was ever called in list context, but that seems unlikely to happen. :-)

10. nov. 2021 12:24:03 Jan Henning Thorsen ***@***.***>:

PR #630, opened 2021-11-26
Add file management, closes #426

### LINKED ISSUE #426: As a user, I would like to manage my uploaded files
There's two reasons why I want this:

* I want to be able to delete files that was uploaded by mistake.
* I might want to copy the URL and resend it to the same / another conversation.

jhthorsen:
Reference: https://github.com/convos-chat/convos/issues/426

### DIFF
--- a/Changes
+++ b/Changes
@@ -1,5 +1,10 @@
 Revision history for perl distribution Convos
 
+6.39 2021-11-27T07:08:00+0900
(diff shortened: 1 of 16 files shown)


jhthorsen:
Done quite a bit of testing to confirm that we don't accidentally delete files that is not visible on a given page. Based on that and the new backend tests, I'm going to merge this now.

PR #644, opened 2021-12-23
As a developer, I want to use features from Perl 5.20

jhthorsen:
There are some features in Perl 5.20 (such as signatures) that will make it more fun to develop Convos. I don't think this is too crazy, since Perl 5.20 was released late 2014.

https://metacpan.org/dist/perl/view/pod/perl5200delta.pod#Core-Enhancements

I'm also considering bumping Convos to v7.00 if this gets merged. Thoughts?

### DIFF
--- a/t/irc-commands.t
+++ b/t/irc-commands.t
@@ -163,7 +163,7 @@ $server->subtest(
 
     $server->server_event_ok('_irc_event_mode')
       ->server_write_ok(
-      ":localhost 367 superman #convos x!*@* superman!~superman@125-12-219-233.rev.home.ne.jp 1577498687\r\n"
+      ":localhost 367 superman #convos x!*\@* superman!~superman\@125-12-219-233.rev.home.ne.jp 1577498687\r\n"
     )->server_write_ok(":localhost 368 superman #convos :End of Channel Ban List\r\n");
     $res = $connection->send_p('#convos', '/mode b')->$wait_success;
     $server->processed_ok;
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -24,7 +24,7 @@ jobs:
     strategy:
       matrix:
         perl-version:
-        - '5.26'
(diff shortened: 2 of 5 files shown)


jhthorsen:
I guess the failing tests in https://github.com/convos-chat/convos/runs/4613160432 means that there has been issues on older perls already, so this PR could actually be an improvement.

stephan48:
As a Reference:
Debian Jessie(LTS until 06/2020) had Perl 5.20.2
Debian Stretch(current LTS until 06/2022) has Perl 5.24.0
Debian Buster(oldstable has) 5.28.0
Bullseye has 5.34.0

My humble opinion is that 5.20.2 is by now plently old enough :)

marcusramberg (approved):

t/irc-commands.t
```diff
@@ -163,7 +163,7 @@ $server->subtest(
 
     $server->server_event_ok('_irc_event_mode')
       ->server_write_ok(
-      ":localhost 367 superman #convos x!*@* superman!~superman@125-12-219-233.rev.home.ne.jp 1577498687\r\n"
+      ":localhost 367 superman #convos x!*\@* superman!~superman\@125-12-219-233.rev.home.ne.jp 1577498687\r\n"
```
stephan48, line 166: Should this change be in this PR?

jhthorsen: Yes, it doesn’t fail in more recent perls

stephan48: ah i see - nvm then :)

jhthorsen:
Oh. This is too bad: I have to choose between signatures and async-await, without bumping to 5.26 😞 Pretty sure I'm going for async-await, since it will make it much easier to add "_p" methods, without having to add more callbacks. I think I have to convert `$self->backend->user` to a promise based method to be able to fix #696.

https://github.com/convos-chat/convos/compare/modern-perl?w=1

PR #702, opened 2022-03-13
Add Convos::Plugin::Auth::Header, closes #696

### LINKED ISSUE #696: Proxy authentication by http header value
**Is your feature request related to a problem? Please describe.**

When self-hosting multiple applications, you really want to have a single point for user management and authentication. It is annoying to login to each and every app seperately.

A pretty simple way to centralize authentication is achieved by deploying apps behind a reverse proxy, and use proxy auth. The proxy handles authentication in some way and sets http headers containing the username that was successfully logged-in. The apps read the headers and associate incoming requests to that user.

**Describe the solution you'd like**

The perfect proxy auth feature for me would work like this:

1. Start convos with additional environment variables:
  * containing the name of the initial admin user (e.g. `admin=admin_user`)
  * enabling proxy auth (e.g. `proxy_auth=true`)
  * setting the key of the http header that contains the username (e.g. `auth_header=X-Authenticated-User`)
2. Configure the reverse proxy to authenticate incoming requests in any way you like.
3. Let the reverse proxy set `X-Authenticated-User` to the authenticated username on every request.
4. Convos treats the requests as if they belong to the appropriate user session.
5. Bonus: if Convos does not know the username, it creates a new user with that name.

**Describe alternatives you've considered**

Other SSO methods like OIDC still require the user to login with each app, even it no credentials are required. It is still an additional step that is unneeded and hurting the user experience.

**Additional context**

I am using Convos for [this product](https://getportal.org/). Since this is a single-user platform, users really should see no login screen at all, not even for SSO.

jhthorsen:
This plugin will allow a user to be registered / logged in by reading a HTTP header. This HTTP header can be set in a reverse proxy, such as nginx.

### DIFF
--- a/lib/Convos/Plugin/Auth/Header.pm
+++ b/lib/Convos/Plugin/Auth/Header.pm
@@ -1,0 +1,90 @@
+package Convos::Plugin::Auth::Header;
+use Mojo::Base 'Convos::Plugin::Auth', -async_await;
+
+has admin_user         => sub { $ENV{CONVOS_ADMIN} // 'admin-email-not-set@example.com' };
+has x_user_header_name => sub { $ENV{CONVOS_AUTH_HEADER} || 'X-Authenticated-User' };
+
+sub register {
+  my ($self, $app, $config) = @_;
+  $self->SUPER::register($app, $config);
+  $self->admin_user($config->{admin_user})                 if defined $config->{admin_user};
+  $self->x_user_header_name($config->{x_user_header_name}) if $config->{x_user_header_name};
+  $app->helper('user.load_p' => sub { $self->_user_load_p(@_) });
+}
+
+async sub _user_load_p {
+  my ($self, $c) = @_;
+  my $core = $c->app->core;
+
+  my $header = $self->x_user_header_name;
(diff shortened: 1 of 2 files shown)


lib/Convos/Plugin/Auth/Header.pm
```diff
@@ -0,0 +1,90 @@
...
+    CONVOS_ADMIN=admin@example.com \
+    CONVOS_AUTH_HEADER=X-User \
+    ./script/convos daemon
+
+C<CONVOS_ADMIN> defaults to C<admin-email-not-set@example.com>, but can be set
+to the Convos admin user's email address to I<disallow> anyone else to register
+as a normal user before the admin has registered/logged in. Setting this
+enviroment variable to an empty string will make the first user the admin user.
+
+C<CONVOS_AUTH_HEADER> must contain the header name that will contain the email
```
max-tet, line 61: Can I use just a string instead of an email address for an user identifier? That would be very usefuld, except if an email is required for some other reason.

max-tet: Alright, thanks!

PR #787, opened 2022-09-09
Fix unnecessary rehashing in `$user->validate_password`

stigtsp:
Crypt::Passphrase::needs_rehash() was passed the plaintext password rather than the hash causing a rehash on every successful login attempt.


jhthorsen (approved):
Thanks! Nice catch 👍

Issue #746, opened 2022-05-31
Hide/fold side panels 


mostlyvirtual:
**Is your feature request related to a problem? Please describe.**
Panels cannot be hidden.

**Describe the solution you'd like**
I would really like the option to hide the panels, when working on a smaller screen, especially the 'Participants' panel, to gain more screen realestate.

jhthorsen:
The panels automatically hides when changing the screen size. Currently this is set both in the javascript and in the stylesheet:

https://github.com/convos-chat/convos/blob/main/assets/App.svelte#L63
https://github.com/convos-chat/convos/blob/main/assets/sass/abstracts/_variables.scss

I'm not sure if I will add buttons to toggle the panels, but if you can provide a screenshot, mockup or reference to another app that shows how this could look, I might consider it.

jhthorsen:
I'll look into it, since I've been wanting to add query params to show/hide parts of the UI for many years, and this functionality is overlapping 👍

PR #821, opened 2023-01-12
Merged settings and participants list

jhthorsen:
One thing that has bugged me for a long time is how it's so difficult to get to settings for a Conversation or Connection.

This PR is a Work In Progress, but I'm opening for discussion. Here is screenshot taken from this PR:

Screenshot 2023-01-12 at 10 05 27

Summary:

* The settings will be collapsed by default, but accessible just by toggling the checkbox.
* The participant groups will be open by default, but can be collapsed.
* No magical cogwheel in the main menu.

Not sure about:

* Should the "Leave" button be inside settings? I'm also considering just adding a "X", since leaving a channel is a non-destructive action. (Nothing gets deleted)
* Do we need a "Save" button for settings?
* Should the top bar stretch from side to side?
* Should the topic box be inside settings or somewhere else? Can it replace the truncated topic in the header? Or should it be a supplement?

TODO:

* All of the above...
* Add number of operators, members, ...
* Add a toggle for hiding the whole sidebar #746
* Store the (un)collapsed state in browser's localStorage.


poVoq:
This translates not very well to mobile use I think.

I agree that currently it is a bit counter intuitive (on desktop at least) that the channel settings button opens a settings menu on the opposite side of the screen over the channel list. But in general having an easily accessible (and touch screen ready) channel settings button seems like not such a bad idea.

I think tiny collapsible harmonicas work somewhat ok with mouse input, but are already annoying when using touchpad and keyboard mostly.

jhthorsen:
This is part of v8.01.

PR #839, opened 2023-08-21
Added Romanian translation, wrote test for it

mostlyvirtual:
In addition to the Romanian translation and the test, I also fixed line 136 in SettingsAccount.svelte


jhthorsen (approved):
I have no idea if the translation is correct or not, so I just have to trust you 😄 

Thanks!

PR #840, opened 2023-08-22
Added collapsible sidebar sections

mostlyvirtual:
This might help with #588 as well.

### DIFF
--- a/assets/components/MainMenu.svelte
+++ b/assets/components/MainMenu.svelte
@@ -20,6 +20,8 @@ let navEl;
 let searchHasFocus = false;
 let visibleLinks = [];
 
+let collapsedStates = JSON.parse(localStorage.getItem('collapsedStates')) || {};
+
 $: filterNav({filter}); // Passing "filter" in to make sure filterNav() is called on change
 $: addConversationLink = '/settings/conversation?connection_id=' + ($user.activeConversation.connection_id || '');
 $: searchQuery = filter.replace(/^\//, '');
@@ -152,9 +154,21 @@ function onSearchKeydown(e) {
 function renderUnread(conversation, max = 60) {
   return conversation.unread > max ? (max + '+') : conversation.unread || 0;
 }
+
+function toggleSection(connection_id) {
+    if (collapsedStates[connection_id]) {
+        collapsedStates[connection_id] = !collapsedStates[connection_id];
+    } else {
+        collapsedStates[connection_id] = true; // default to collapsed if not already in the object
+    }
+
+    // Save to localStorage
+    localStorage.setItem('collapsedStates', JSON.stringify(collapsedStates));
+}
+
 </script>
 
-<style>
+<style lang="scss">
 .header-wrapper {
   background: var(--sidebar-left-bg);
   position: sticky;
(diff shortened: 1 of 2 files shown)


assets/components/MainMenu.svelte
```diff
@@ -20,6 +20,8 @@ let navEl;
 let searchHasFocus = false;
 let visibleLinks = [];
 
+let collapsedStates = JSON.parse(localStorage.getItem('collapsedStates')) || {};
```
jhthorsen, line 23: What happens to JSON.parse() when collapsedStates is not defined in localStorage?

assets/components/MainMenu.svelte
```diff
@@ -152,9 +154,21 @@ function onSearchKeydown(e) {
...
+
+function toggleSection(connection_id) {
+    if (collapsedStates[connection_id]) {
+        collapsedStates[connection_id] = !collapsedStates[connection_id];
+    } else {
+        collapsedStates[connection_id] = true; // default to collapsed if not already in the object
+    }
+
+    // Save to localStorage
+    localStorage.setItem('collapsedStates', JSON.stringify(collapsedStates));
```
jhthorsen, line 166: Please change the indentation for consistency.

assets/components/MainMenu.svelte
```diff
@@ -152,9 +154,21 @@ function onSearchKeydown(e) {
 function renderUnread(conversation, max = 60) {
   return conversation.unread > max ? (max + '+') : conversation.unread || 0;
 }
+
+function toggleSection(connection_id) {
+    if (collapsedStates[connection_id]) {
+        collapsedStates[connection_id] = !collapsedStates[connection_id];
+    } else {
+        collapsedStates[connection_id] = true; // default to collapsed if not already in the object
+    }
```
jhthorsen, line 163: Don't think you need all the if/else logic. Just do:

```javascript
function toggleSection(connection_id) {
  collapsedStates[connection_id] = !collapsedStates[connection_id];
  localStorage.setItem('collapsedStates', JSON.stringify(collapsedStates));
}
```

Or do you want the sections to be collapsed as default..? I'm not very fond of hiding information from new users.

assets/components/MainMenu.svelte
```diff
@@ -152,9 +154,21 @@ function onSearchKeydown(e) {
...
+
+    // Save to localStorage
+    localStorage.setItem('collapsedStates', JSON.stringify(collapsedStates));
+}
+
 </script>
 
-<style>
+<style lang="scss">
```
jhthorsen, line 171: Why is this changed?

assets/components/MainMenu.svelte
```diff
@@ -184,80 +198,97 @@ function renderUnread(conversation, max = 60) {
...
+        <b class="badge-connection" hidden="{!connection.unread}" on:click={e => {
+          toggleSection(connection.connection_id);
+          }}>{renderUnread(connection)}
+        </b>
+        <!-- Here's an icon indicating collapsed/expanded state -->
+        <div class="chevron"><Icon
+          name={collapsedStates[connection.connection_id] ? 'chevron-right' : 'chevron-down'}
+          on:click={e => {
+          toggleSection(connection.connection_id);}}/>
+        </div>
```
jhthorsen, line 221: This section has a bit strange formatting. Here is my suggestion:

```html
<b class="badge-connection" hidden="{!connection.unread}" on:click={() => toggleSection(connection.connection_id)}}>{renderUnread(connection)}</b>
<!-- Here's an icon indicating collapsed/expanded state -->
<div class="chevron">
  <Icon name={collapsedStates[connection.connection_id] ? 'chevron-right' : 'chevron-down'}
    on:click={e => toggleSection(connection.connection_id)}/>
</div>

File: assets/components/MainMenu.svelte [hunk 6]
```diff
@@ -184,80 +198,97 @@ function renderUnread(conversation, max = 60) {
...
-    {#if $user.is('admin')}
-      <Link href="/settings">
-        <Icon name="tools"/>
-        <span>{$l('Settings')}</span>
+
+    <h3 class="account-toggle-plus" on:click={() => toggleSection($user.email)}>
+      <span>{$user.email || $l('Account')}</span>
+      <Icon name={collapsedStates[$user.email] ? 'plus' : 'minus'} />
+    </h3>
+    <div class={collapsedStates[$user.email] ? 'collapsedSection' : ''}>
```
jhthorsen, line 244: Not a fan of camelCase in classnames. But either way, I think you can just use the "hidden" class, which already exists. I would even suggest using that on the links above...

jhthorsen (changes requested):
Some things that I think needs tuning, but overall I think we'll add the functionality.

Just one big thing: Right now, I don't think you're able to search for "hidden" conversations. I think it makes sense that the conversations appear if searched for.

mostlyvirtual:
Thanks for taking the time to review this. I'll look into each requested change and reply in-line. Probably next week.

assets/components/MainMenu.svelte
mostlyvirtual: I think the || {} fallback should take care of that. If collapsedStates is not defined in localStorage, then it will be initialized as an empty object.

assets/components/MainMenu.svelte
mostlyvirtual: I think you're right. If there is no cookie set, then all sections should be expanded, else collapse the ones that need to be collapsed. I've updated the code with your suggestion.

assets/components/MainMenu.svelte
mostlyvirtual: To be honest, I think it might've been from some earlier testing and hacking. I've removed that change for now, and will run through the test cases to see if I can replicate the reason why I put it in in the first place.

assets/components/MainMenu.svelte
mostlyvirtual:
```suggestion
    <div class="collapsedSection" hidden={!collapsedStates[$user.email]}>
```

assets/components/MainMenu.svelte
mostlyvirtual:
```suggestion
        <b class="badge-connection" hidden="{!connection.unread}" on:click={e => { toggleSection(connection.connection_id); }} on:keydown={e => { if (e.keyCode === 13) { toggleSection(connection.connection_id); } }}>{renderUnread(connection)}</b>
        <!-- Here's an icon indicating collapsed/expanded state -->
        <div class="chevron">
          <Icon name={collapsedStates[connection.connection_id] ? 'chevron-right' : 'chevron-down'}
          on:click={e => { toggleSection(connection.connection_id);}}/>
        </div>
```

assets/components/MainMenu.svelte
mostlyvirtual: Actually, that's wrong, it should be:
    <div hidden={!collapsedStates[$user.email]}>

mostlyvirtual:
```suggestion
    <div class={collapsedStates[$user.email] && !filter.length ? 'collapsed-section' : ''}>
```

I'm making this suggestion, as I tried using the hidden class, but then the filter/search for things like "Files" or "Settings" wouldn't work anymore.

mostlyvirtual:
> Just one big thing: Right now, I don't think you're able to search for "hidden" conversations. I think it makes sense that the conversations appear if searched for.

Actually, this looks like it was an easy and clean fix after about 2 hours of trying ugly solutions :)

It involved adding filter.length to this line for the conversations/channels:
`      {#if !collapsedStates[connection.connection_id] || filter.length}
`

And filter.length to this line for the account section:

`    <div class={collapsedStates[$user.email] && !filter.length ? 'collapsedSection' : ''}>
`

assets/components/MainMenu.svelte
jhthorsen: I just tried it myself, and it seems like you're right. At least in Firefox. I'm not sure why though, since it doesn't look like parse() should be able to take null. Please change it just to be sure:

```javascript
let collapsedStates = JSON.parse(localStorage.getItem('collapsedStates') || '{}');
```

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/parse

mostlyvirtual: I'm not entirely sure what the difference is tbh, but from what I can tell it also doesn't hurt, so I implemented your snippet.

marcusramberg (approved):
lgtm!

jhthorsen:
So busy today -- forgot to say thanks 😄

mostlyvirtual:
> So busy today -- forgot to say thanks 😄

My pleasure. I feel you on the busy part, don't worry. Glad we reached a point where we could merge. Thank you.

PR #863, opened 2023-11-09
gha: build docker image for linux/arm/v7 (RPi 3)

cadubentzen:
I tried to run convos-chat on my RPi 3B+, but found out that there are no docker images for linux/arm/v7.

Built the image via buildx locally on my laptop and it's running fine on the RPi 3B+.

Depends on https://github.com/convos-chat/docker-base/pull/24

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -38,7 +38,7 @@ jobs:
           context: ./
           file: ./Dockerfile
           push: true
-          platforms: linux/amd64,linux/arm64
+          platforms: linux/amd64,linux/arm64,linux/arm/v7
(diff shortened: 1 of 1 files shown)


jhthorsen:
Thanks! Looks good to me 👍

jhthorsen:
Don't have time to make a new release just now, but the "alpha" image should be built as expected soon.

cadubentzen:
@jhthorsen I think the build will fail if https://github.com/convos-chat/docker-base/pull/24 isn't merged as well.

jhthorsen:
Hm... Strange that it didn't fail and also strange that it's listed as "unknown/unknown ". Got any ideas?

https://github.com/convos-chat/convos/pkgs/container/convos/146201501?tag=alpha

I tried to rebuild it, after I merged docker-base as well.

